PR #1, opened 2021-08-25
[FFM-1260-61] Project structure, authentication, fetching initial data

enver-bisevac:
* Authentication
* Fetch flags and segments


milos85vasic-harness (approved):

src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      // prepare cache for storing segments
+      console.log(response.data);
+    }).catch((error: Error) => {
+      console.error('Error loading segments', error)
+    })
+
+    await Promise.all([flags, segments]);
+  }
+
+  async run() {
```
knagurski, line 72:
```suggestion
  async run(): Promise<void> {
```

src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  boolVariation(identifier: string, target: Target, defaultValue: false): boolean {
+    console.log(identifier, target)
+    return defaultValue;
+  }
+}
+
+
+const client = new CfClient('sdk key');
+client.boolVariation('test', null, false);
```
knagurski, line 88: Did you mean to leave in debug stuff?

src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  async fetchData() {
+    const flags = this.api.getFeatureConfig(this.environment, {
+      params: {
+        'cluster': this.cluster
+      }
+    }).then(response => {
+      // prepare cache for storing flags
+      console.log(response.data)
+    }).catch((error: Error) => {
+      console.error('Error loading flags', error)
```
knagurski, line 56: Instead of just logging this to the console, it might be worth throwing it so it can be caught in the user's code.

knagurski (changes requested):
The lack of a default on the constructor could be a problem

src/index.ts
enver-bisevac: Yes, for the moment, later to setup inner example project

src/index.ts
enver-bisevac: Done

PR #4, opened 2021-09-13
[FFM-1263] flag rules evaluations, repository with persistent

enver-bisevac:
Repository patternt with cache support
Flag rules evaluations

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,5 +17,7 @@
     "plugin:jest/recommended",
     "prettier"
   ],
-  "rules": {}
+  "rules": {
+    "@typescript-eslint/no-unused-vars": [2, { "argsIgnorePattern": "^_" }]
+  }
 }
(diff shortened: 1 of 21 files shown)


davejohnston (commented):
This looks good to me (though I'm not experienced in Node).   The repository pattern and evaluator seem like a good approach here.

chrisblakely01 (commented):
Looks pretty good! Do you think we should put some unit tests in?

enver-bisevac:
> Looks pretty good! Do you think we should put some unit tests in?

Yes, after we all agree this is the way we should go. Evaluator should be 100% covered, that is for us most important part.

chrisblakely01:
> > Looks pretty good! Do you think we should put some unit tests in?
> 
> Yes, after we all agree this is the way we should go. Evaluator should be 100% covered, that is for us most important part.

Cool, sounds good to me

chrisblakely01 (approved):

PR #6, opened 2021-09-15
[FFM-1264] segment evaluator

enver-bisevac:


src/evaluator.ts
```diff
@@ -70,12 +75,19 @@ export class Evaluator {
...
+  ): string | undefined {
+    let variation = '';
     for (const _var of distribution.variations) {
+      variation = _var.variation;
       if (this.isEnabled(target, distribution.bucketBy, _var.weight))
         return _var.variation;
     }
-    return undefined;
+    return this.isEnabled(target, distribution.bucketBy, ONE_HUNDRED)
+      ? variation
```
knagurski, line 90: Why does the final variation get precedence?

src/evaluator.ts
```diff
@@ -92,6 +105,72 @@ export class Evaluator {
     }
   }
 
+  private isTargetInList(target: Target, targets: Target[]): boolean {
+    return (
+      targets.find((elem) => elem.identifier === target.identifier) != undefined
```
knagurski, line 110: You can simplify with `includes`
```suggestion
      return targets.includes(target)
```

knagurski: Actually, since the vanilla includes is pretty readable, maybe this method isn't required

src/evaluator.ts
```diff
@@ -92,6 +105,72 @@ export class Evaluator {
...
+
+  private convertApiTargetsToEvalTargets(targets: ApiTarget[]): Target[] {
+    return targets.map(
+      (elem) =>
+        ({
+          identifier: elem.identifier,
+          name: elem.name,
+          anonymous: elem.anonymous,
+          attributes: elem.attributes,
+        } as Target),
```
knagurski, line 122: Do you need this type assertion? The return type on the method should be sufficient

src/evaluator.ts
```diff
@@ -92,6 +105,72 @@ export class Evaluator {
...
+
+        // Should Target be included via segment rules
+        if (segment.rules) {
+          if (this.evaluateClauses(segment.rules, target)) {
+            // log.debug(
+            //   "Target %s included in segment %s via rules\n",
+            //   target.getName(), segment.getName());
+            return true;
+          }
+        }
```
knagurski, line 168: Might as well join this conditions in a single `if`
```suggestion
        if (segment.rules && this.evaluateClauses(segment.rules, target)) {
          // log.debug(
          //   "Target %s included in segment %s via rules\n",
          //   target.getName(), segment.getName());
          return true;
        }
```

enver-bisevac: make sense

src/evaluator.ts
enver-bisevac: compiler complains about mismatched types

src/evaluator.ts
```diff
@@ -154,27 +234,61 @@ export class Evaluator {
...
-      if (fc.defaultServe.distribution) {
-        const identifier = this.getKeyName(fc.defaultServe.distribution, target);
-        variation = this.findVariation(fc, identifier);
-      } else {
-        variation = this.findVariation(fc, fc.defaultServe.variation);
+  private evaluateVariationMap(
+    variationToTargetMap: VariationMap[],
+    target: Target,
+  ): string | undefined {
+    if (!variationToTargetMap) return undefined;
```
knagurski, line 232: No need to change, but there's no need to return `undefined` specifically. Not returning, returning without a value and returning `undefined` all do the same thing.

src/evaluator.ts
```diff
@@ -154,27 +234,61 @@ export class Evaluator {
...
+
+    for (const variationMap of variationToTargetMap) {
+      // find target
+      const targetMap = variationMap.targets?.find(
+        (elem) => elem.identifier === target.identifier,
+      );
+      if (targetMap) return variationMap.variation;
```
knagurski, line 239: You can simplify a wee bit
```suggestion
      if (variationMap.targets?.includes(target)) {
        return variationMap.variation;
      }
```

src/evaluator.ts
```diff
@@ -154,27 +234,61 @@ export class Evaluator {
...
+      if (
+        segmentIdentifiers &&
+        this.isTargetIncludedOrExcludedInSegment(segmentIdentifiers, target)
+      ) {
+        return variationMap.variation;
       }
     }
-    return variation
+
+    return undefined;
```
knagurski, line 251: Again, no real need to specifically return `undefined`. Not returning will do the same thing. No need to change

src/evaluator.ts
```diff
@@ -154,27 +234,61 @@ export class Evaluator {
...
+    let variation = this.evaluateVariationMap(fc.variationToTargetMap, target);
+    if (variation) return this.findVariation(fc.variations, variation);
+
+    // not found in variationMaps proceed with rules processor
+    variation = this.evaluateRules(fc.rules, target);
+    if (variation) return this.findVariation(fc.variations, variation);
+
+    // not found in rules proceed with default serve
+    variation = fc.defaultServe.variation;
+    // default serve variation is undefined so continue with distribution
```
knagurski: It doesn't appear that you are testing whether the default serve variation is undefined

src/evaluator.ts
enver-bisevac: compiler complains

src/index.ts
```diff
@@ -95,11 +95,35 @@ export class CfClient {
   boolVariation(
     identifier: string,
     target: Target,
-    defaultValue: false,
+    defaultValue = false,
```
knagurski, line 98: Good catch. I make this error ALL the time in TS

knagurski (approved):
Looks great. Just a couple of observations and nit-picks

src/evaluator.ts
enver-bisevac: the reason I did in this way: targets is TargetMap type and target is Target type

PR #7, opened 2021-09-23
[FFM-1266] metrics processor done

enver-bisevac:


src/metrics.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  const _summarize = (): Metrics | unknown => {
+    if (!data) {
+      log.debug('No metrics data!');
+      return;
+    }
+    const targetData: TargetData[] = [];
+    const metricsData: MetricsData[] = [];
+
+    for (const [, event] of Object.entries(data)) {
```
knagurski, line 91: You can simplify a little bit with `Object.values` since you don't need the key.
```suggestion
    for (const event of Object.values(data)) {
```

src/metrics.ts
```diff
@@ -0,0 +1,157 @@
...
+      if (event.target && !event.target.anonymous) {
+        const targetAttributes: KeyValue[] = [];
+        if (event.target.attributes)
+          for (const [key, value] of Object.entries(event.target.attributes))
+            targetAttributes.push({ key, value: value as string });
```
knagurski, line 96: Try to avoid dropping the brackets, it would only take a small change to break this is unexpected ways.

src/metrics.ts
knagurski: This looks like a decent use-case for a `reduce`

src/metrics.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+    return {
+      targetData: targetData,
+      metricsData: metricsData,
+    };
+  };
+
+  const _send = (): void => {
+    const metrics: Metrics = _summarize();
+    api.postMetrics(environment, metrics, {
```
knagurski, line 132: AS `metrics` might be `unknown` at this point, do you want to post it off without checking?

knagurski (approved):
Looks good. Just a few suggestions 👍🏻

milos85vasic-harness (approved):

src/metrics.ts
```diff
@@ -0,0 +1,157 @@
...
+      event.count = 1;
+      data[key] = event;
+    }
+  };
+
+  const _formatKey = (event: AnalyticsEvent): string => {
+    const feature = event.featureConfig.feature,
+      variation = event.variation.identifier,
+      value = event.variation.value,
+      target = GLOBAL_TARGET;
```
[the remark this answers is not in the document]
enver-bisevac: make sense

milos85vasic-harness (approved):

PR #13, opened 2021-09-30
[MAINT] retry mechanism using axios-retry

enver-bisevac:


src/streaming.ts
```diff
@@ -124,7 +124,7 @@ export class StreamProcessor {
     return this.eventSource.readyState == StreamProcessor.CONNECTED;
   }
 
-  private stop(): void {
+  stop(): void {
```
knagurski, line 127: Any reason you'd want this to be public, but not the equivalent function in `polling.ts`? Just curious

knagurski (approved):
Looks great. That package looks like it does exactly what we want with only a tiny bit of code... well found.

src/streaming.ts
enver-bisevac: Good catch :), I was doing a test and I needed to stop it and then forgot to return to private modifier

PR #14, opened 2021-09-30
[MAINT] minor fixes

enver-bisevac:
- Set data to repository only if new version of object
- Sort rules by priority
- Added test case

### DIFF
--- a/__tests__/usecases/rules_priority.json
+++ b/__tests__/usecases/rules_priority.json
@@ -1,0 +1,78 @@
+{
+    "flag": {
+      "defaultServe": { "distribution": null, "variation": "true" },
+      "environment": "demoenv",
+      "feature": "AlfaTest",
+      "kind": "boolean",
+      "offVariation": "false",
+      "prerequisites": [],
+      "project": "FeatureFlagsDemo",
+      "rules": [
+        {
+          "ruleId": "",
+          "priority": 1000,
+          "clauses": [
+            {
+              "id": "",
+              "attribute": "identifier",
+              "op": "equal",
+              "values": ["harness"]
+            }
+          ],
+          "serve": {
+            "distribution": null,
+            "variation": "true"
+          }
+        },
+        {
+            "ruleId": "",
+            "priority": 999,
+            "clauses": [
+              {
+                "id": "",
+                "attribute": "identifier",
+                "op": "equal",
+                "values": ["harness"]
+              }
+            ],
+            "serve": {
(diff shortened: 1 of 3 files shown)


__tests__/usecases/rules_priority.json
```diff
@@ -0,0 +1,78 @@
+{
```
knagurski, line 1: Not necessary, but usually a good idea to run things like this through prettier for consistency

enver-bisevac: Done

src/repository.ts
```diff
@@ -59,38 +65,48 @@ export class StorageRepository implements Repository {
...
         this.cache.set(segmentKey, segment);
       }
       return segment;
     }
     return undefined;
   }
 
+  private checkFlagVersion(key: string, flag: FeatureConfig): boolean {
+    const oldFlag = this.getFlag(key, false); // dont set cacheable, we are just checking the version
+    return !oldFlag || !oldFlag.version || oldFlag.version < flag?.version;
```
knagurski, line 102: You _could_ simplify, but it's fine as-is
```suggestion
    return !oldFlag?.version || oldFlag.version < flag?.version;
```

src/evaluator.ts
```diff
@@ -216,6 +216,11 @@ export class Evaluator {
     if (!target || !rules) {
       return undefined;
     }
+
+    rules.sort((a: ServingRule, b: ServingRule) =>
+      a.priority > b.priority ? 1 : b.priority > a.priority ? -1 : 0,
```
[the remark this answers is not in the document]
enver-bisevac: Make sense

src/repository.ts
```diff
@@ -59,38 +65,48 @@ export class StorageRepository implements Repository {
...
-      if (segment) {
+      if (segment && cacheable) {
         this.cache.set(segmentKey, segment);
       }
       return segment;
     }
     return undefined;
   }
 
+  private checkFlagVersion(key: string, flag: FeatureConfig): boolean {
```
knagurski, line 100: Just personal preference, but I prefer a the function name to describe what it'll give me back. In the case of a boolean like this, something like `isFlagOutdated`

src/repository.ts
enver-bisevac: make sense :)

knagurski (approved):
I don't like this.


I love it! ❤️

PR #15, opened 2021-09-30
[MAINT] readme modified

enver-bisevac:


README.md
```diff
@@ -15,6 +15,93 @@ To sign up, https://app.harness.io/auth/#/signup/
...
 
+### Setup
+
+```npm install @harness/ff-nodejs-server-sdk```
+
+### Import the package (CommonJS)
+
+```
+const pkg = require('ff-nodejs-server-sdk');
+const { Client } = pkg;
```
knagurski, line 26: Might as well combine these. Either with object access or destructuring.
```suggestion
const { Client } = require('ff-nodejs-server-sdk');
```

```suggestion
const Client = require('ff-nodejs-server-sdk').Client;
```

enver-bisevac: Done

milos85vasic-harness (approved):

PR #16, opened 2021-10-01
[MAINT] Prerequisite checks done

enver-bisevac:


davejohnston (approved):
This looks good to me.

src/evaluator.ts
```diff
@@ -289,22 +298,92 @@ export class Evaluator {
...
+    }
+
+    return undefined;
+  }
+
+  boolVariation(
+    identifier: string,
+    target: Target,
+    defaultValue = false,
+    callback: Callback = undefined,
```
knagurski, line 378: If this is `undefined` by default, should `callback` be optional?
```suggestion
    callback?: Callback,
```

knagurski (approved):
Looks good, just a wee thing

src/evaluator.ts
enver-bisevac: Done

PR #17, opened 2021-10-02
[MAINT] Async repository

enver-bisevac:


milos85vasic-harness (approved):

README.md
```diff
@@ -56,7 +56,7 @@ const target = {
 
 ### Evaluate the flag with default value set to false
 ```
```
knagurski, line 58: Not 100% required, but GitHub will give better highlighting if you tell it what language your code is in.
```suggestion
```typescript
```

README.md
```diff
@@ -66,24 +66,25 @@ client.close();
 
 ### Avaialable public methods
 ```
-const value = client.boolVariation('test', target, false);
-const value = client.stringVariation('test', target, 'BLACK');
-const value = client.numberVariation('test', target, 1);
-const value = client.jsonVariation('test', target, {});
+function boolVariation(identifier: string, target: Target, defaultValue = true): Promise<boolean>;
```
knagurski, line 69: At some point we should make these links to documentation on the specific functions and how to use them. Another time though

README.md
```diff
@@ -66,24 +66,25 @@ client.close();
...
-  baseUrl: string;
-  eventsUrl: string;
-  pollInterval: number;
-  eventsSyncInterval: number;
-  enableStream: boolean;
-  enableAnalytics: boolean;
-  cache: KeyValueStore;
-  store: KeyValueStore;
-  logger: Logger;
+baseUrl: string;
```
knagurski, line 79: As above, we make the user aware they exist, but don't explain them.

example/index_cf.mjs
```diff
@@ -2,7 +2,7 @@ import CfClient from 'ff-nodejs-server-sdk';
 
 CfClient.init('1c100d25-4c3f-487b-b198-3b3d01df5794');
 
-setInterval(() => {
-    const value = CfClient.boolVariation('test', null, false);
+setInterval(async() => {
+    const value = await CfClient.boolVariation('test', null, false);
     console.log("Evaluation for flag test and target none: ", value);
 }, 10000);
```
knagurski, line 8: Don't worry about it for now, but it looks like prettier isn't running automatically on save for you. There's a mix of single and double quotes and a missing newline at the end of the file. Probably worth configuring prettier to run on save for `.mjs` files as well.

src/client.ts
```diff
@@ -115,11 +115,11 @@ export default class Client {
     log.info('finished setting up processors');
   }
 
-  boolVariation(
+  async boolVariation(
```
knagurski, line 118: It doesn't look like this function is `async`; you're not using `await`, but rather just returning the result from `this.evaluator.boolVariation`. Your options are to strip the `async` or `await` the `boolVariation` call.

src/client.ts
```diff
@@ -132,11 +132,11 @@ export default class Client {
     );
   }
 
-  stringVariation(
+  async stringVariation(
```
knagurski, line 135: Same as above

src/evaluator.ts
```diff
@@ -176,7 +177,10 @@ export class Evaluator {
     return false;
   }
 
-  private evaluateClause(clause: Clause, target: Target): boolean {
+  private async evaluateClause(
```
knagurski, line 180: You probably want to `await` any async calls in the `switch` statement as well

src/index.ts
```diff
@@ -1,42 +1,56 @@
...
     }
   },
-  boolVariation: function (
+  boolVariation: async function (
     identifier: string,
     target: Target,
     defaultValue = false,
-  ): boolean {
+  ): Promise<boolean> {
     return this.instance.boolVariation(identifier, target, defaultValue);
```
knagurski, line 30: Does `this` work in this context? i.e. not inside a class? Might be worth checking it out. The `this.instance = new Client` on line 22 might be actually pointing to the global scope, not this exported object, so these `this.instance.*` calls might only be working because they too are referencing the global scope.

If you change one of these functions to use the method syntax, you may be able to prove that one way or the other.

```typescript
export default {
  // ...
  
  async boolVariation(
    identifier: string,
    target: Target,
    defaultValue = false,
  ): Promise<boolean> {
    return this.instance.boolVariation(identifier, target, defaultValue);
  },

  // ...
}
```

src/index.ts
```diff
@@ -1,42 +1,56 @@
...
   },
-  stringVariation: function (
+  stringVariation: async function (
     identifier: string,
     target: Target,
     defaultValue = '',
-  ): string {
+  ): Promise<string> {
     return this.stringVariation(identifier, target, defaultValue);
```
knagurski, line 37: This is a recursive call. Was it supposed to be...
```suggestion
    return this.instance.stringVariation(identifier, target, defaultValue);
```

src/index.ts
```diff
@@ -1,42 +1,56 @@
...
+};
 export default {
   instance: undefined,
   init: function (sdkKey: string, options: Options): void {
     if (!this.instance) {
       this.instance = new Client(sdkKey, options);
     }
   },
-  boolVariation: function (
+  boolVariation: async function (
```
knagurski, line 25: Since you're not using `await`, you don't need `async`, unless you want to add `await` to the `this.instance.boolVariation` call

src/repository.ts
```diff
@@ -1,78 +1,84 @@
...
 
 export class StorageRepository implements Repository {
   private cache: KeyValueStore;
-  private store: KeyValueStore;
+  private store: AsyncKeyValueStore;
 
-  constructor(cache: KeyValueStore, store?: KeyValueStore) {
+  constructor(cache: KeyValueStore, store?: AsyncKeyValueStore) {
     if (!cache) {
       throw new Error('Cache is required argument and connot be undefined');
```
knagurski, line 20:
```suggestion
      throw new Error('Cache is required argument and cannot be undefined');
```

src/repository.ts
```diff
@@ -1,78 +1,84 @@
...
     this.store = store;
   }
 
-  setFlag(identifier: string, fc: FeatureConfig): void {
+  async setFlag(identifier: string, fc: FeatureConfig): Promise<void> {
     const flagKey = this.formatFlagKey(identifier);
-    if (this.isFlagOutdated(flagKey, fc)) {
-      return;
+    if (await this.isFlagOutdated(flagKey, fc)) {
+      return undefined;
```
knagurski, line 29: Is `undefined` required? Your function is defined as returning `void`.

src/repository.ts
```diff
@@ -1,78 +1,84 @@
...
     }
     if (this.store) {
-      this.store.set(flagKey, fc);
-      this.cache.del(flagKey);
+      this.store.set(flagKey, fc).then((_value) => {
+        this.cache.del(flagKey);
+      });
```
knagurski, line 34: Nothing wrong with this, but you can simplify with an `await`
```suggestion
      await this.store.set(flagKey, fc).then((_value) => {
      this.cache.del(flagKey);
```

src/repository.ts
```diff
@@ -1,78 +1,84 @@
...
 
-  deleteFlag(identifier: string): void {
+  deleteFlag(identifier: string): Promise<void> {
     const flagKey = this.formatFlagKey(identifier);
     if (this.store) {
       this.store.del(flagKey);
     }
     this.cache.del(flagKey);
+    return undefined;
   }
```
knagurski, line 63: I'm not seeing how this function returns a `Promise`.

src/repository.ts
```diff
@@ -1,78 +1,84 @@
...
 
-  deleteSegment(identifier: string): void {
+  deleteSegment(identifier: string): Promise<void> {
```
knagurski, line 65: As above

src/store.ts
```diff
@@ -1,26 +1,23 @@
...
-    }
-
-}
\ No newline at end of file
+  set(key: string, value: unknown): Promise<true> {
+    return this.keyv.set(key, value);
+  }
+  get(key: string): Promise<unknown> {
+    return this.keyv.get(key);
+  }
```
knagurski, line 19: It looks like `keyv.get` supports generics, might be worth adding that to your code as well. Then you can do things like `store.get<MyType>(x)` instead of `store.get(x) as MyType`. Just a suggestion, don't worry about it.

knagurski (changes requested):
Some potential issues with `src/index.ts` and some other bits and pieces to look at, but looks like it's really coming together. Since this is breaking change (i.e. the user _will have to_ update their code to use it), make sure to mark this as `v2.0.0`

src/index.ts
knagurski: Never mind, it was driving my IDE crazy

knagurski:
Oh, and it looks like there are about 778 vulnerabilities in the dependencies. It would be worth running `npm audit` to reveal them and `npm audit fix` to fix them. Just make sure everything still works after the fixes.

README.md
enver-bisevac: make sense

src/client.ts
enver-bisevac: make sense

src/evaluator.ts
enver-bisevac: only this is promise
```
return this.isTargetIncludedOrExcludedInSegment(clause.values, target);
```

milos85vasic-harness (approved):

src/repository.ts
```diff
@@ -1,78 +1,82 @@
...
       return;
     }
     if (this.store) {
-      this.store.set(flagKey, fc);
+      await this.store.set(flagKey, fc);
       this.cache.del(flagKey);
     } else {
       this.cache.set(flagKey, fc);
     }
+    return;
```
knagurski, line 37: Since you're not returning early or returning anything, you don't need trailing `return`'s like this.

src/repository.ts
```diff
@@ -81,14 +85,14 @@ export class StorageRepository implements Repository {
...
-  getSegment(identifier: string, cacheable = true): Segment {
+  async getSegment(identifier: string, cacheable = true): Promise<Segment> {
     const segmentKey = this.formatSegmentKey(identifier);
     let segment = this.cache.get(segmentKey) as Segment;
     if (segment) {
       return segment;
     }
     if (this.store) {
-      segment = this.store.get(segmentKey) as Segment;
+      segment = await this.store.get<Segment>(segmentKey);
```
knagurski, line 95: Definitely not something to think about now, but at some point we need to go through all of the SDK's and replace `Segment` with `TargetGroup` to make sure we have consistency across UI and SDK.

src/store.ts
```diff
@@ -1,26 +1,23 @@
...
-    get(key: string): unknown {
-        return this.keyv.get(key);
-    }
-    del(key: string): void {
-        this.keyv.delete(key);
-    }
-
-}
\ No newline at end of file
+  set<T>(key: string, value: T): Promise<true> {
```
knagurski, line 14: Probably don't need the generic here as it doesn't really have any effect.

src/repository.ts
```diff
@@ -1,78 +1,82 @@
...
+      await this.store.set(segmentKey, segment);
       this.cache.del(segmentKey);
     } else {
       this.cache.set(segmentKey, segment);
     }
+    return;
   }
 
-  deleteFlag(identifier: string): void {
+  async deleteFlag(identifier: string): Promise<void> {
```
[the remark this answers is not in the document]
enver-bisevac:
```
if (this.store) {
      this.store.del(flagKey);
}
```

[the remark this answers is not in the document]
knagurski: As is, it won't actually wait for that to complete before the promise resolves, so `await` it.

src/repository.ts
enver-bisevac: Done

src/repository.ts
enver-bisevac: make sense

src/repository.ts
```diff
@@ -1,78 +1,78 @@
...
     } else {
       this.cache.set(segmentKey, segment);
     }
   }
 
-  deleteFlag(identifier: string): void {
+  async deleteFlag(identifier: string): Promise<void> {
     const flagKey = this.formatFlagKey(identifier);
     if (this.store) {
       this.store.del(flagKey);
```
knagurski, line 55: This needs to be `await`ed

src/repository.ts
```diff
@@ -1,78 +1,78 @@
...
     }
     this.cache.del(flagKey);
   }
 
-  deleteSegment(identifier: string): void {
+  async deleteSegment(identifier: string): Promise<void> {
     const segmentKey = this.formatSegmentKey(identifier);
     if (this.store) {
       this.store.del(segmentKey);
```
knagurski, line 63: This too

src/repository.ts
enver-bisevac: Done

src/repository.ts
enver-bisevac: Done

milos85vasic-harness (approved):

knagurski (approved):
Nice work 👍🏻

PR #21, opened 2021-10-14
[FFM-1501] initial waitForInit implementation

enver-bisevac:
This PR introduce waitFor init before any usage of public methods

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,17 +18,19 @@ To sign up, https://app.harness.io/auth/#/signup/
 
 ### Setup
 
-`npm install @harnessio/ff-nodejs-server-sdk`
+```shell
+npm install @harnessio/ff-nodejs-server-sdk
+```
 
 ### Import the package (CommonJS)
 
-```
+```javascript
 const { Client } = require('ff-nodejs-server-sdk');
 ```
 
 ### Import the package (ES modules)
 
-```
+```typescript
 import { Client } from 'ff-nodejs-server-sdk';
 ```
 
@@ -42,9 +44,9 @@ const client = new Client('your server type SDK key');
 
 Advanced initialization can be done using options
 
-```
+```typescript
 const client = new Client('your server type SDK key', {
-  enableStream: false
+  enableStream: false,
 });
 ```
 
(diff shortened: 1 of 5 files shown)


milos85vasic-harness (approved):

README.md
```diff
@@ -34,18 +35,21 @@ import { Client } from 'ff-nodejs-server-sdk';
 ### Initialize
 
 This is the most simple way to initialize SDK using only a server type key
+
 ```
```
knagurski, line 39: Try to add the language type here so GitHub can be more intelligent about its highlighting
```suggestion
```typescript
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ff-nodejs-server-sdk",
-  "version": "1.0.1",
+  "version": "1.1.1",
```
knagurski, line 3: Since we introduced a breaking change by completely changing the API of the SDK to be asynchronous across the board, this really aught to be at `2.0.0` by now if we're sticking to SemVer

README.md
```diff
@@ -17,7 +18,7 @@ To sign up, https://app.harness.io/auth/#/signup/
 
 ### Setup
 
-```npm install @harness/ff-nodejs-server-sdk```
+`npm install @harness/ff-nodejs-server-sdk`
```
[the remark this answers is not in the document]
enver-bisevac: make sense

package.json
knagurski: By changing the returns of all the variation function to be promises, everything is broken. Code written for `v1.0.0` will not work with `v1.1.0`

knagurski: Also, this should only be `v1.1.1` if it _only_ fixes bugs in `v1.1.0`, if there are any breaking changes (e.g. changes in function signatures/return types), then it should be `v2.0.0` and any new features or changes that do not break the existing signature should be `v1.1.0`

enver-bisevac: The new thing is we introduce waitForInit so I think it is a requested feature from customer :), So that is the reason for 1.1.0 version

PR #25, opened 2021-10-18
[FFM-1507] metrics event target fixed, minor changes

enver-bisevac:
- When metric event target is undefined in ESM module throws an exception
- Minor changes in Query interface signature

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -157,7 +157,7 @@ export const MetricsProcessor = (
         },
         {
           key: TARGET_ATTRIBUTE,
-          value: event.target.identifier,
+          value: event?.target?.identifier ?? null,
         },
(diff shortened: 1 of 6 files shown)


milos85vasic-harness (approved):

src/metrics.ts
```diff
@@ -157,7 +157,7 @@ export const MetricsProcessor = (
         },
         {
           key: TARGET_ATTRIBUTE,
-          value: event.target.identifier,
+          value: event?.target?.identifier ?? null,
```
knagurski, line 160: Just checking that `undefined` isn't allowed here and that it should either have a value or be `null`

knagurski (approved):
Looks good 👍🏻

PR #29, opened 2022-01-19
[FFM-1551] integrate ff-test-cases repo as submodule, refactor evaluateClause

enver-bisevac:
* fix progress rollout
* refactor evaluate clause to compare on string types
* integrate ff-test-cases


chrisblakely01 (approved):
LGTM 👍

jest.config.js
```diff
@@ -1,20 +1,11 @@
...
-  ],
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
   testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(ts|js)x?$',
   coverageDirectory: 'coverage',
-  collectCoverageFrom: [
-    'src/**/*.{ts,tsx,js,jsx}',
-    '!src/**/*.d.ts',
-  ],
+  coveragePathIgnorePatterns: ['src/openapi/'],
+  collectCoverageFrom: ['src/**/*.{ts,tsx,js,jsx}', '!src/**/*.d.ts'],
```
knagurski, line 10: Not a problem, but specifying we want coverage from tsx and jsx files is unnecessary since we're unlikely to be writing React or Preact components in the node SDK.
```suggestion
  collectCoverageFrom: ['src/**/*.{ts,js}', '!src/**/*.d.ts'],
```

jest.config.js
```diff
@@ -1,20 +1,11 @@
...
   },
-  moduleFileExtensions: [
-    "ts",
-    "tsx",
-    "js",
-    "jsx",
-    "json",
-    "node",
-  ],
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
```
knagurski, line 6: Here are well, we can drop tsx and jsx
```suggestion
  moduleFileExtensions: ['ts', 'js', 'json', 'node'],
```

jest.config.js
enver-bisevac: 👍

jest.config.js
enver-bisevac: 👍

src/evaluator.ts
```diff
@@ -181,26 +157,38 @@ export class Evaluator {
     clause: Clause,
     target: Target,
   ): Promise<boolean> {
-    if (!clause.op) {
+    if (!clause?.op || !clause?.values || clause.values.length === 0) {
```
knagurski, line 160: You can probably shorten this a bit
```suggestion
    if (!clause?.op || !clause?.values.length) {
```

knagurski (approved):
Looks good 👍🏻 Only a few completely optional suggestions

src/evaluator.ts
enver-bisevac: ```if (!clause?.op || !clause?.values.length)``` if values is undefined?: ```if (!clause?.op || !clause?.values?.length)```

PR #38, opened 2022-04-06
[FFM-2583] on close stream starts pollprocessor again

enver-bisevac:
When the stream processor is stopped the poll processor is started again.


src/__tests__/client.test.ts
```diff
@@ -0,0 +1,22 @@
+import Client from '../client';
+import { PollingProcessor } from '../polling';
+
+jest.mock('../openapi/api');
+
+it('check close client', async () => {
```
knagurski, line 6: Totally personal preference, but a test's name should read like a requirement. If you're using `it`, then it makes sense that the test name should start with `should` and describe the expected outcome.

In this case, maybe something like...
```suggestion
it('should close the client when the close method is called', async () => {
```

src/__tests__/client.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  const start = jest.spyOn(PollingProcessor.prototype, 'start');
+  const close = jest.spyOn(PollingProcessor.prototype, 'close');
+
+  // when
+  const client = new Client('some key', {
+    enableAnalytics: false,
+  });
+  await client.waitForInitialization();
+
+  client.close();
```
knagurski, line 17: I haven't checked it locally, but this like might trigger a warning about an action being taken outside of an `act` block. If not, don't worry about it 👍🏻

src/client.ts
```diff
@@ -113,7 +114,9 @@ export default class Client {
     });
 
     this.eventBus.on(StreamEvent.DISCONNECTED, () => {
-      this.pollProcessor.start();
+      if (!this.closing) {
+        this.pollProcessor.start();
+      }
```
knagurski, line 119: Isn't it crazy how many times a bug can come down to a single missed check? Nice find and fix 👍🏻

knagurski (approved):
Looks good, just a few suggestions

PR #51, opened 2022-07-14
fix: [FFM-4047]: Ensure async calls are awaited

knagurski:
The PR fixes an issue whereby evaluations of Target Group rules were not awaited properly, leading to the wrong variation being served.

Refs [FFM-4047](https://harness.atlassian.net/browse/FFM-4047)


bmjen (approved):
This looks good. The test failures are unrelated and have been failing for a while. Will open a ticket to get those fixed separately.

PR #54, opened 2022-07-26
[MAINT] fix for target evaluation

enver-bisevac:
Problem:
When evaluate flag with target it always return default value

Fix:
Target group evaluation is async function and it was invoked without `await` so promise returned instead of real value.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.2.5",
+  "version": "1.2.6",
```
knagurski, line 3: Looks like `package-lock.json` doesn't have this version. You probably just ned to run `npm i`

knagurski (approved):
Apart from the version number in the `package-lock.json` this is great work. Well spotted.

PR #55, opened 2022-08-18
FFM-4331 - Change evaluation logic for Target Groups

stephenmcconkey:
**What**
Changes evaluation logic for target groups, so that the logic works as OR's rather than AND's... I.e. only one of the conditions must be true 

**Why**
[FFM-4331](https://harness.atlassian.net/browse/FFM-4331)
[FFM-4219](https://harness.atlassian.net/browse/FFM-4219)


enver-bisevac (approved):
LGTM

PR #60, opened 2022-11-18
[FFM-5387]: Update Test Cases

davejohnston:
Update the test-cases submodule.
This pulls in over 70 new tests for the evaluator.

The evaluator test had to be updated to understand the newer format.  We also had to modify how the JSON results are confirmed.


chrisblakely01 (approved):
LGTM 👍

PR #62, opened 2023-02-24
FFM-4204 Fallback to polling if stream has an error

erdirowlands:
# What
Previously, the SDK would not fall back to polling if there was an error in the stream. This was caused by a number of issues
- The SDK would not signal the client to do this on `StreamEvent.ERROR` events
- The `eventsource` library had a number of issues and so we have forked it and put our own fixes in there. See [PR](https://github.com/EventSource/eventsource/pull/317) for details. The SDK uses the new #harness.io/event NPM package which is pending publication to NPM. ** BUILD WILL FAIL UNTIL PACKAGE PUBLISHED **

# Testing
Manual. Extensive.
- See testing on the above PR. In all of those scenarios, the SDK would fall back to polling.


package.json
```diff
@@ -64,7 +63,7 @@
   "dependencies": {
     "axios": "^0.22.0",
     "axios-retry": "^3.2.0",
-    "eventsource": "^1.1.2",
+    "@harness/event": "file:../eventsource",
```
erdirowlands, line 66: This will be updated once the package is published to NPM. Feel free to mark the PR as requested changes until this is done, along with any other required changes.

andybharness (approved):

PR #63, opened 2023-03-02
FFM-7002 Tweak polling to fall back on `onretry` event as well as `onerror` event

erdirowlands:
# What

- I've just made the eventsource library no longer emit errors if it is retrying, so this change makes the SDK fallback to polling on retry events. It will still fallback to polling on error events, just that the eventsource library won't emit error events unless it encounters a non-retryable errors.
- Bumps eventsource lib version

# Testing
Manual testing


examples/example/package-lock.json
```diff
@@ -1,115 +0,0 @@
-{
```
erdirowlands, line 1: I'll regenerate the package-lock files once the new version of es is up and can run npm i again

andybharness (approved):

examples/getting_started/package.json
```diff
@@ -3,6 +3,6 @@
   "version": "1.0.0",
   "main": "index.js",
   "dependencies": {
-    "@harnessio/ff-nodejs-server-sdk": "file:../.."
+    "@harnessio/ff-nodejs-server-sdk": "1.2.11"
```
knagurski, line 6: Any specific reason for changing this?

erdirowlands: No. Thanks, fixing.

PR #65, opened 2023-03-23
fix: [FFM-7289]: Defer initialization resolve

knagurski:
This PR defers the resolving of the `waitForInitialization` Promise to address a race conditional.


erdirowlands (approved):
Nice.

PR #69, opened 2023-04-13
FFM-7412 - Bump ff-eventsource dependency to fix socket leak

andybharness:
FFM-7412 - Bump ff-eventsource dependency to fix socket leak

What
This updates the Harness fork of the eventsource lib to use an updated bug fixed version: If we're already retrying a stream endpoint, avoid scheduling another retry. Instead let the existing retry run its course.

Why
We're seeing the SDK endlessly spin with retries when a stream endpoint drops.

Testing
Tested manually with a custom proxy to drop the stream endpoint and reproduce the spin

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.2.15",
+  "version": "1.2.16",
   "description": "Feature flags SDK for NodeJS environments",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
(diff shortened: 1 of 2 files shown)


bmjen (approved):

bmjen:
Approved, but need to release eventsource and rerun tests before this one can go through.

erdirowlands (changes requested):
The ff-eventsource version that this PR bumps cause the library not to do any retries. We can't merge this until that is sorted.

Sticking in "Request changes" for now.

erdirowlands:
Closing as @andybharness has implemented SSE without using an external library in 

https://github.com/harness/ff-nodejs-server-sdk/pull/70

PR #70, opened 2023-04-19
FFM-7412 - Reduce complexitiy of SSE code

andybharness:
[FFM-7412] Potential memory leak on SSE retries

**What**
Remove eventsource dependency to reduce code complexity and create an easier to maintain implementation

***Why***
We're seeing the eventsource/retry packages spin and cause a lot of retries to happen at once, causing an adverse effect on system resources.

***Testing***
Manual testing with custom proxies to simulate network failures.

[FFM-7412]: https://harness.atlassian.net/browse/FFM-7412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/streaming.ts
```diff
@@ -46,46 +52,96 @@ export class StreamProcessor {
...
-          'API-Key': this.apiKey,
-        },
-      },
-    );
-
-    eventSource.onopen = (event: MessageEvent) => {
-      this.log.debug('Stream connected', event);
+    this.log.info('Starting new StreamProcessor');
+
+    const url = `${this.options.baseUrl}/stream?cluster=${this.cluster}`
```
bmjen, line 65: Does this section need to exist? it looks like the openapi generated client already has a helper to call the `stream` endpoint. But that also begs the question of why this wasn't used instead of `eventsource`
https://github.com/harness/ff-nodejs-server-sdk/blob/3388147168527e77abcc84bb2f29d4e62e0f7814/src/openapi/api.ts#L1078

andybharness: Without knowing the historical context of the code, my guess is the openapi spec includes this and generates the relevant code but we weren't using it anyhow (instead we had eventsource), possibility due to headers needing added or due to the SSE socket needing to live for a long time, unlike regular rest requests. Some of the openapi generators can be quiet buggy too (for other language).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.2.15",
+  "version": "1.2.16",
```
knagurski, line 3: Looks like the lock file wasn't updated. Can you run `npm i` to update it and commit that?

src/streaming.ts
```diff
@@ -1,11 +1,14 @@
...
 import EventEmitter from 'events';
 import { AxiosPromise } from 'axios';
 import { ClientApi, FeatureConfig, Segment } from './openapi';
 import { StreamEvent, Options, StreamMsg } from './types';
 import { Repository } from './repository';
 import { ConsoleLog } from './log';
 
+import * as https from 'https';
+import * as http from 'http';
+import { RequestOptions } from 'https';
```
knagurski, line 10: I think you can simplify this as just...
```suggestion
import https, { RequestOptions } from 'https';
import http from 'http';
```

It's also good practice to import things that are from the platform first (in this case before `EventEmitter`), but deffo not required

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
     this.eventBus = eventBus;
     this.repository = repository;
     this.log = this.options.logger;
+
+    const minDelay = 5000;
+    const maxDelay = 10000;
+    this.retryAttempt = 1;
+    this.retryDelayMs = Math.floor(
+      Math.random() * (maxDelay - minDelay + 1) + minDelay,
+    );
```
knagurski, line 61: I'm confused by this... since `minDelay` and `maxDelay` are both consts, then this equates to `Math.random() * 5001 + 5000` is that intentional? Seems pretty arbitrary and will always come up with a number between 5.001s and 10.001s.

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
+        Accept: 'text/event-stream',
+        Authorization: `Bearer ${this.jwtToken}`,
+        'API-Key': this.apiKey,
       },
-    );
+    };
 
-    eventSource.onopen = (event: MessageEvent) => {
-      this.log.debug('Stream connected', event);
+    const connected = () => {
```
knagurski, line 78: It's usually good practice to prefix handlers with `on` to set them apart. Or use the suffix `Handler`. Or both.
```suggestion
    const onConnectedHandler = () => {
```

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
 
-    eventSource.onretrying = (event: MessageEvent) => {
-      this.log.error('Stream is trying to reconnect', event);
-      // Only inform the client once that we're retrying
-      if (!this.isRetrying) {
-        this.isRetrying = true
-        this.eventBus.emit(StreamEvent.RETRYING);
-      }
-    };
+    const failed = (msg) => {
```
knagurski, line 84: If we know the type, we should add it here.
```suggestion
    const failed = (msg: string): void => {
```

src/streaming.ts
andybharness: yeah it increments a random delay between 5 and 10 seconds up until a max of 60 seconds. 5-10 seconds was chosen to prevent it reaching 60s too fast. not the most sophisticated back off strategy but simple to implement with some level of randomness. the expression is a bit confusing because i couldn't find a method to give me a random number between two ranges in node.

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
+  ): void {
+    if (this.readyState === StreamProcessor.CONNECTED) {
+      this.log.debug('SSE already connected, skip retry');
+      return;
+    }
+
+    const isSecure = url.startsWith('https:');
+    this.log.debug('SSE HTTP start request', url);
+
+    (isSecure ? https : http)
```
knagurski, line 121: I was going to suggest we not support `http`, but instead insist on a secure connection, but I suppose there is a use-case for communication between the SDK and the proxy within the same vpc.

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
+          failed('HTTP code ' + res.statusCode);
+          return;
+        }
+
+        connected();
+
+        res
+          .on('data', (data) => {
+            this.processData(data);
+          })
```
knagurski, line 135: Super minor and deffo not a blocker, but you can simplify a wee bit
```suggestion
          .on('data', this.processData)
```

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
+
+    const isSecure = url.startsWith('https:');
+    this.log.debug('SSE HTTP start request', url);
+
+    (isSecure ? https : http)
+      .request(url, options, (res) => {
+        this.log.debug('SSE got HTTP response code', res.statusCode);
+
+        if (res.statusCode !== 200) {
+          failed('HTTP code ' + res.statusCode);
```
knagurski, line 126: Not an issue, but most other places we use template strings over string concat.
```suggestion
          failed(`HTTP code ${res.statusCode}`);
```

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
+        connected();
+
+        res
+          .on('data', (data) => {
+            this.processData(data);
+          })
+          .on('close', () => {
+            failed('SSE stream closed');
+          });
+        //.on('end', ()=> { failed('SSE stream ended'); })
```
knagurski, line 139: Do we need this line?

src/streaming.ts
knagurski: I know what you mean, I miss `rand(min, max)` in PHP. But anyway, you can safely remove the `+1` part and it'll be fine unless you really want it to be 5.001s and 10.001s

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
     this.eventBus = eventBus;
     this.repository = repository;
     this.log = this.options.logger;
+
+    const minDelay = 5000;
+    const maxDelay = 10000;
+    this.retryAttempt = 1;
```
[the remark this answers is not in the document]
knagurski: Really this whole thing could do with being rewritten in more of a functional approach. But not today.

src/streaming.ts
andybharness: It's also useful local test proxies running in http/plaintext, meaning you don't have to setup a bunch of local test certs

src/streaming.ts
```diff
@@ -46,46 +51,116 @@ export class StreamProcessor {
...
+
+  private processData(data: any): void {
+    const lines = data.toString().split(/\r?\n/);
+    lines.forEach((line) => this.processLine(line));
+  }
+
+  private processLine(line: string): void {
+    if (line.startsWith('data:')) {
+      this.log.debug('SSE GOT:', line.substring(5));
+      const msg = JSON.parse(line.substring(5));
```
knagurski, line 163: Minor thing, but just to keep things readable, it's best to keep a bit of white space before conditions.
```suggestion
      const msg = JSON.parse(line.substring(5));

```

src/streaming.ts
```diff
@@ -46,46 +52,116 @@ export class StreamProcessor {
...
-    eventSource.onretrying = (event: MessageEvent) => {
-      this.log.error('Stream is trying to reconnect', event);
-      // Only inform the client once that we're retrying
-      if (!this.isRetrying) {
-        this.isRetrying = true
-        this.eventBus.emit(StreamEvent.RETRYING);
-      }
-    };
+    const failed = (msg) => {
+      const delayMs = this.getRandomRetryDelayMs();
```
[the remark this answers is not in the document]
knagurski: The logic is exactly the same... you're just incrementing after with the current implementation. Incrementing before would the exact same operations, but less confusion. There won't be a `* 0` in either case.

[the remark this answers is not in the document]
andybharness: i see what you mean now, a get function that modifies the state. sure i can move that out

src/polling.ts
```diff
@@ -52,6 +53,13 @@ export class PollingProcessor {
       this.timeout = setTimeout(() => this.poll(), sleepFor);
     };
 
+    if (this.lastPollTime > (Date.now() - this.options.pollInterval)) {
+      this.log.info(`Last poll was ${(Date.now()-this.lastPollTime)/1000} seconds ago, skipping flag refresh`);
```
knagurski, line 57: This will almost certainly wind up being a non-whole number. We're best passing it through either `Math.round` or `Math.ceil` to make it more readable.

```suggestion
      this.log.info(`Last poll was ${(Date.now()-this.lastPollTime)/1000} seconds ago, skipping flag refresh`);
```

src/polling.ts
```diff
@@ -20,6 +20,7 @@ export class PollingProcessor {
   private eventBus: EventEmitter;
   private timeout: NodeJS.Timeout;
   private log: ConsoleLog;
+  private lastPollTime  = 0;
```
knagurski, line 23: Looks like there's an extra space here. Remember to use `prettier` on save for these kinda things.

src/streaming.ts
```diff
@@ -46,45 +52,113 @@ export class StreamProcessor {
...
-        this.eventBus.emit(StreamEvent.RETRYING);
-      }
-    };
+    const onFailed = (msg: string) => {
+      this.retryAttempt += 1;
 
-    eventSource.onerror = (event: MessageEvent) => {
-      this.log.error('Unrecoverable error with stream, not retrying to connect: ', event);
-      this.eventBus.emit(StreamEvent.ERROR, event);
+      const delayMs = this.getRandomRetryDelayMs() * this.retryAttempt;
```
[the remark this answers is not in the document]
andybharness: i spotted that after i submitted earlier, that will be fixed in a subsequent push

src/polling.ts
andybharness: I'll run npm format here and check in separately (as to not cause too much noise in the diff)

bmjen:
Force Push was to remove my change to the github actions workflow.

bmjen (approved):

src/streaming.ts
```diff
@@ -46,45 +52,113 @@ export class StreamProcessor {
...
+    this.log.debug('SSE HTTP start request', url);
+
+    (isSecure ? https : http)
+      .request(url, options, (res) => {
+        this.log.debug('SSE got HTTP response code', res.statusCode);
+
+        if (res.statusCode !== 200) {
+          onFailed(`HTTP code ${res.statusCode}`);
+          return;
+        }
```
knagurski, line 129: This is a wee bit too simplistic since it will work fine with any 2xx or probably 3xx. I think what was there before was a check for any 4xx or 5xx response to trigger the error. It'll probably be fine for now 👍🏻

knagurski (approved):
Looks the part 👍🏻

src/streaming.ts
andybharness: i can add 4xx/5xx here, it'll be a simple one line change.

looks like the eventsource lib had this

`  function isRetryableError (error) {
    if (error) {
      if (error.status) {
        var status = error.status
        return status === 500 || status === 502 || status === 503 || status === 504
      }
    }
    // Always retry on IO Errors
    return true
  }`

andybharness: just seeing this now, i went for

`if (res.statusCode >= 400 && res.statusCode <= 599)`

PR #72, opened 2023-06-05
FFM-8116 Stop SSE when `Client.close()` is called

erdirowlands:
# What
Adds functionality to abort the SSE stream if `Client.close()` is called. 

# Testing
Manually called `close()` and the SSE connection aborts and does not retry connecting

### DIFF
--- a/src/streaming.ts
+++ b/src/streaming.ts
@@ -7,7 +7,6 @@ import { ConsoleLog } from './log';
 
 import https, { RequestOptions } from 'https';
 import http from 'http';
-import { AbortController } from 'abort-controller';
 
 
 type FetchFunction = (
@@ -30,6 +29,7 @@ export class StreamProcessor {
   private readonly retryDelayMs: number;
 
   private options: Options;
+  private abortController: AbortController;
   private eventBus: EventEmitter;
   private readyState: number;
   private log: ConsoleLog;
@@ -67,6 +67,9 @@ export class StreamProcessor {
 
     const url = `${this.options.baseUrl}/stream?cluster=${this.cluster}`;
 
+
+
+
     const options = {
       headers: {
         'Cache-Control': 'no-cache',
@@ -74,6 +77,7 @@ export class StreamProcessor {
         Authorization: `Bearer ${this.jwtToken}`,
         'API-Key': this.apiKey,
       },
+      // signal: { signal }
     };
 
     const onConnected = () => {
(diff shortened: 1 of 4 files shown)


src/streaming.ts
```diff
@@ -115,12 +122,20 @@ export class StreamProcessor {
...
+    if (this.abortController) {
+      this.abortController.abort();
+    }
+    this.abortController = new AbortController();
+    const {signal} = this.abortController;
 
     const isSecure = url.startsWith('https:');
     this.log.debug('SSE HTTP start request', url);
 
+    const appendedOptions = {...options, signal};
```
erdirowlands, line 135: We have to create a new object from scratch which includes the original http options + the signal to abort. Otherwise, the type checking fails, so it is merely to satisfy the typescript linter.

src/streaming.ts
```diff
@@ -1,13 +1,14 @@
...
-import { AxiosPromise } from 'axios';
-import { ClientApi, FeatureConfig, Segment } from './openapi';
-import { StreamEvent, Options, StreamMsg } from './types';
-import { Repository } from './repository';
-import { ConsoleLog } from './log';
+import {AxiosPromise} from 'axios';
+import {ClientApi, FeatureConfig, Segment} from './openapi';
+import {StreamEvent, Options, StreamMsg} from './types';
+import {Repository} from './repository';
+import {ConsoleLog} from './log';
```
knagurski, line 6: Did you turn off Prettier?

src/streaming.ts
```diff
@@ -17,6 +18,7 @@ type FetchFunction = (
 export class StreamProcessor {
   static readonly CONNECTED = 1;
   static readonly RETRYING = 2;
+  static readonly CLOSED = 3;
```
knagurski, line 21: Don't change in this PR, but this really aught to be an enum, not readonly static vars

src/streaming.ts
erdirowlands: Ooops. I am not sure - I might not have it installed and set as the formatter of choice. I will fix now.

src/streaming.ts
```diff
@@ -212,14 +228,18 @@ export class StreamProcessor {
     return this.readyState === StreamProcessor.CONNECTED;
   }
 
-  stop(): void {
-    this.log.info('Stopping StreamProcessor');
-    this.eventBus.emit(StreamEvent.DISCONNECTED);
-  }
```
knagurski, line 218: Looks like this was a public function. Is this change potentially breaking and require a `2.0.0`?

src/streaming.ts
erdirowlands: That is really ugly formatting. Headdesk.

knagurski: I copy-paste a config between projects and Prettier is the first thing I ever add. I hate having to think about formatting.

erdirowlands: Donezel washington now

andybharness (approved):

src/streaming.ts
```diff
@@ -115,12 +119,20 @@ export class StreamProcessor {
...
+    if (this.abortController) {
+      this.abortController.abort();
+    }
+    this.abortController = new AbortController();
+    const { signal } = this.abortController;
 
     const isSecure = url.startsWith('https:');
     this.log.debug('SSE HTTP start request', url);
 
+    const appendedOptions = { ...options, signal };
```
knagurski, line 132: We might have an issue with this approach... officially we tell customers we support Node 12+ with this SDK, but the abort signal and `AbortController` was introduced in Node 15 https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility. We may need to either feature detect or change our support matrix to be 15+. The reason you were getting a typescript issue was that the `@types/node` is set as `14`, so bumping that to `15` will allow `signal` in the options

knagurski (approved):
Looks okay 👍

PR #73, opened 2023-06-22
FFM-8289 Stop SDK crashing on auth failures / SDK Codes / JWT Validation

erdirowlands:
# What
This PR focuses on three changes

1. If the SDK fails authentication, then the SDK will not attempt to start any services (which results in a crash) and will serve default variations to users. 

2. Adds a list of SDK codes we are rolling out across SDKs - see https://harness.atlassian.net/wiki/spaces/FFM/pages/21356052573/SDK+Codes 

3. We saw an issue reported by a customer that the `environment` claim was missing from the auth JWT. This adds some validation and logging for that rare scenario. 

# Testing
Manual: 

1. SDK initialises and serves the expected flag variations 
2. SDK fails to initialise due to auth error or falsey API key and continues to to serve defaults
3. SDK Codes are logged correctly - I've included unit tests for these just to make sure all that code is tested.
4. Valid JWT claims don't cause the SDK to stop auth

### DIFF
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -27,7 +27,6 @@ describe('SDK Tests', () => {
       sdkCodes.debugStreamEventReceived(logger);
       sdkCodes.infoStreamStopped(logger);
       sdkCodes.infoMetricsSuccess(logger);
-      sdkCodes.infoMetricsTargetExceeded(logger);
       sdkCodes.infoMetricsThreadExited(logger);
       sdkCodes.debugEvalSuccess(
         'dummy result',
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -1,4 +1,4 @@
-import { Logger } from '../log';
+import type { Logger } from '../log';
 import * as sdkCodes from '../sdk_codes';
 
 describe('SDK Tests', () => {
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -1,0 +1,20 @@
+import { ConsoleLog, Logger } from "../log";
+import * as sdkCodes from '../sdk_codes';
+
+
+describe('SDK Tests', () => {
+  let logger: Logger;
+
+  beforeEach(() => {
+    logger = new ConsoleLog()
+  });
+
+  test('Run all sdk_code functions without raising exceptions', () => {
+    expect(() => {
+      sdkCodes.infoSDKInitOK(logger);
+      sdkCodes.infoSDKCloseSuccess(logger);
+      sdkCodes.infoMetricsThreadStarted(5000, logger);
+      // Call other functions here
+    }).not.toThrow();
+  });
+});
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -1,11 +1,17 @@
-import { ConsoleLog, Logger } from "../log";
+import { Logger } from "../log";
 import * as sdkCodes from '../sdk_codes';
 
 describe('SDK Tests', () => {
   let logger: Logger;
 
   beforeEach(() => {
-    logger = new ConsoleLog()
+    logger = {
+      trace: jest.fn(),
+      debug: jest.fn(),
+      info: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+    };
   });
 
   test('Run all sdk_code functions without raising exceptions', () => {
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -20,7 +20,6 @@ describe('SDK Tests', () => {
       sdkCodes.infoSDKCloseSuccess(logger);
       sdkCodes.infoMetricsThreadStarted(5000, logger);
       sdkCodes.infoPollStarted(60, logger)
-      sdkCodes.infoSDKInitWaiting( logger)
       sdkCodes.infoSDKStartClose( logger)
       sdkCodes.infoSDKAuthOK( logger)
       sdkCodes.infoPollingStopped("Dummy reason", logger)
@@ -30,7 +29,7 @@ describe('SDK Tests', () => {
       sdkCodes.infoMetricsSuccess( logger)
       sdkCodes.infoMetricsTargetExceeded( logger)
       sdkCodes.infoMetricsThreadExited( logger)
-      sdkCodes.debugEvalSuccess( logger)
+      sdkCodes.debugEvalSuccess("dummy result", "dummy identifier", { name: "dummy", identifier: "dummy"}, logger)
       sdkCodes.warnAuthFailedSrvDefaults( logger)
       sdkCodes.warnMissingSDKKey( logger)
       sdkCodes.warnFailedInitAuthError( logger)
@@ -39,7 +38,7 @@ describe('SDK Tests', () => {
       sdkCodes.warnStreamDisconnected("dummy reason", logger)
       sdkCodes.warnStreamRetrying(4, logger)
       sdkCodes.warnPostMetricsFailed("dummy error", logger)
-      sdkCodes.warnDefaultVariationServed("flag", "target", "default value", logger)
+      sdkCodes.warnDefaultVariationServed("flag", { name: "dummy", identifier: "dummy"}, "default value", logger)
       // Call other functions here
     }).not.toThrow();
   });
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -1,7 +1,7 @@
 import type { Logger } from '../log';
 import * as sdkCodes from '../sdk_codes';
 
-describe('SDK Tests', () => {
+describe('SDK Codes', () => {
   const logger: Logger = {
     trace: jest.fn(),
     debug: jest.fn(),
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -2,17 +2,13 @@ import { Logger } from '../log';
 import * as sdkCodes from '../sdk_codes';
 
 describe('SDK Tests', () => {
-  let logger: Logger;
-
-  beforeEach(() => {
-    logger = {
-      trace: jest.fn(),
-      debug: jest.fn(),
-      info: jest.fn(),
-      warn: jest.fn(),
-      error: jest.fn(),
-    };
-  });
+  const logger: Logger = {
+    trace: jest.fn(),
+    debug: jest.fn(),
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+  };
 
   test('Run all sdk_code functions without raising exceptions', () => {
     expect(() => {
--- a/src/__tests__/sdk_codes.test.ts
+++ b/src/__tests__/sdk_codes.test.ts
@@ -1,4 +1,4 @@
-import { Logger } from "../log";
+import { Logger } from '../log';
 import * as sdkCodes from '../sdk_codes';
 
 describe('SDK Tests', () => {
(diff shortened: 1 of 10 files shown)


src/metrics.ts
```diff
@@ -177,13 +183,18 @@ export const MetricsProcessor = (
   };
 
   const _send = (): void => {
+    if (closed) {
```
erdirowlands, line 186: I noticed that if any evaluations were made after the SDK had been closed, we'd still process and send them. This stops that from happening.

erdirowlands: It won't stop them being queued - but this is the easier win.

src/__tests__/sdk_codes.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  beforeEach(() => {
+    logger = {
+      trace: jest.fn(),
+      debug: jest.fn(),
+      info: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+    };
+  });
```
knagurski, line 15: No need to put this in a `beforeEach` since you don't assert on it and it's never modified.
```suggestion
  const logger: Logger = {
    trace: jest.fn(),
    debug: jest.fn(),
    info: jest.fn(),
    warn: jest.fn(),
    error: jest.fn(),
  };
```

src/__tests__/sdk_codes.test.ts
```diff
@@ -0,0 +1,53 @@
+import { Logger } from '../log';
+import * as sdkCodes from '../sdk_codes';
+
+describe('SDK Tests', () => {
```
knagurski, line 4: SDK Tests is extremely vague. What are we actually testing here? What is the "Unit" we're exercising? The word "Tests" is superfluous since we know these are tests. Use just the name of the unit.

src/__tests__/sdk_codes.test.ts
```diff
@@ -0,0 +1,53 @@
+import { Logger } from '../log';
```
knagurski, line 1: You're only using the type, so no need for a full import
```suggestion
import type { Logger } from '../log';
```

src/__tests__/sdk_codes.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      sdkCodes.warnStreamRetrying(4, logger);
+      sdkCodes.warnPostMetricsFailed('dummy error', logger);
+      sdkCodes.warnDefaultVariationServed(
+        'flag',
+        { name: 'dummy', identifier: 'dummy' },
+        'default value',
+        logger,
+      );
+    }).not.toThrow();
+  });
```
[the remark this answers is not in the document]
erdirowlands: Excellent. It reminds me of a test generator in Erlang. Thanks.

src/__tests__/sdk_codes.test.ts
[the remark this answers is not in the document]
erdirowlands: What they should do is log to the console. I don't think we need to assert that `console.log` is called once for each of these logging functions. 
This suite is merely to get the entire surface of these logging functions executed and proven to not throw. You can argue that that is not the purpose of unit tests.

src/__tests__/sdk_codes.test.ts
erdirowlands: Wow nice... committing this change

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+    return 'Unknown SDK code';
+  }
+}
+
+export function sdkErrMsg(errorCode: number, appendText = ''): string {
+  return `SDKCODE:${errorCode}: ${getSDKCodeMessage(errorCode)} ${appendText}`;
+}
+
+export function warnMissingSDKKey(logger: Logger): void {
+  logger.warn(sdkErrMsg(1002, ''));
```
knagurski, line 53: `appendText` is optional, so you can omit the blank string
```suggestion
  logger.warn(sdkErrMsg(1002));
```

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+  6000: 'Evaluation successful: ',
+  6001: 'Evaluation Failed, returning default variation: ',
+  // SDK_METRICS_7xxx
+  7000: 'Metrics thread started with request interval:',
+  7001: 'Metrics stopped',
+  7002: 'Posting metrics failed, reason:',
+  7003: 'Metrics posted successfully',
+};
+
+export function getSDKCodeMessage(key: number): string {
```
knagurski, line 40: Probably shouldn't export this function since it's an implementation detail
```suggestion
function getSDKCodeMessage(key: number): string {
```

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+  if (key in sdkCodes) {
+    return sdkCodes[key];
+  } else {
+    return 'Unknown SDK code';
+  }
+}
+
+export function sdkErrMsg(errorCode: number, appendText = ''): string {
```
knagurski, line 48: Same here, this is an internal implementation detail
```suggestion
function getSdkErrMsg(errorCode: number, appendText = ''): string {
```

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
+import { Logger } from './log';
+import { Target } from './types';
+
+interface SDKCodeMessages {
+  [key: number]: string;
+}
+
+const sdkCodes: SDKCodeMessages = {
```
knagurski, line 8: Probably no need for this interface since it can be replaced with `Record`
```suggestion
const sdkCodes: Record<number, string> = {
```

knagurski: You could also consider just making this into an enum

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+export function warnMissingSDKKey(logger: Logger): void {
+  logger.warn(sdkErrMsg(1002, ''));
+}
+
+export function infoPollStarted(durationSec: number, logger: Logger): void {
+  logger.info(sdkErrMsg(4000, `${durationSec * 1000}`));
+}
+
+export function infoSDKInitOK(logger: Logger): void {
+  logger.info(sdkErrMsg(1000, ''));
```
knagurski, line 61: Same for all of these, the second arg is optional, so just let it default to an empty string.

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+export function infoSDKCloseSuccess(logger: Logger): void {
+  logger.info(sdkErrMsg(3001, ''));
+}
+
+export function infoSDKAuthOK(logger: Logger): void {
+  logger.info(sdkErrMsg(2000, ''));
+}
+
+export function infoPollingStopped(logger: Logger): void {
+  logger.info(sdkErrMsg(4001));
```
knagurski, line 77: This one is gonna look like a bug when it outputs `Polling stopped, reason:` with no reason

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const sdkCodes: SDKCodeMessages = {
+  // SDK_INIT_1xxx
+  1000: 'The SDK has successfully initialized',
+  1001: 'The SDK has failed to initialize due to an authentication error - defaults will be served',
+  1002: 'The SDK has failed to initialize due to a missing or empty API key - defaults will be served',
+  // SDK_AUTH_2xxx
+  2000: 'Authentication was successful',
+  2001: 'Authentication failed with a non-recoverable error',
+  2002: 'Authentication attempt',
```
knagurski, line 16: It's super not obvious here that this is an error state

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+  logger.warn(sdkErrMsg(2003, ''));
+}
+
+export function warnAuthRetrying(
+  attempt: number,
+  error: string,
+  logger: Logger,
+): void {
+  logger.warn(
+    sdkErrMsg(2002, `attempt ${attempt}, got error: ${error}, Retrying ...`),
```
knagurski, line 141: This will look a bit odd when it outputs `Authentication attempt attempt 3, got error: something, Retrying...`

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export function infoStreamStopped(logger: Logger): void {
+  logger.info(sdkErrMsg(5004, ''));
+}
+
+export function infoMetricsThreadStarted(
+  interval: number,
+  logger: Logger,
+): void {
+  logger.info(sdkErrMsg(7000, `${interval}`));
```
knagurski, line 96: No real need to put this into a string. Maybe better to just allow `sdkErrMsg` to take either strings or numbers.

src/sdk_codes.ts
```diff
@@ -0,0 +1,171 @@
...
+  // SDK_CLOSE_3xxx
+  3000: 'Closing SDK',
+  3001: 'SDK Closed successfully',
+  // SDK_POLL_4xxx
+  4000: 'Polling started, intervalMs:',
+  4001: 'Polling stopped, reason:',
+  // SDK_STREAM_5xxx
+  5000: 'SSE stream successfully connected',
+  5001: 'SSE stream disconnected, reason:',
+  5002: 'SSE event received: ',
```
knagurski, line 27: Looks like this should have details of the event, but it'll just have 2 spaces instead.

src/evaluator.ts
```diff
@@ -448,6 +454,12 @@ export class Evaluator {
       callback,
     );
     if (variation) {
+      debugEvalSuccess(
+        `${JSON.stringify(variation.value)}`,
```
knagurski, line 458: Won't this already be a JSON encoded string?

src/metrics.ts
```diff
@@ -177,13 +183,18 @@ export const MetricsProcessor = (
...
+      return;
+    }
+
     const metrics: Metrics = _summarize();
     if (metrics) {
       log.debug('Start sending metrics data');
       api
         .postMetrics(environment, cluster, metrics)
         .then((response) => {
           log.debug('Metrics server returns: ', response.status);
```
knagurski, line 196: Should all these older messages be killed off now that we have the standardised ones?

src/__tests__/sdk_codes.test.ts
erdirowlands: Renamed to `SDK Codes`

src/__tests__/sdk_codes.test.ts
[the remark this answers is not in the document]
erdirowlands: Done!

src/sdk_codes.ts
erdirowlands: Awesome!

src/sdk_codes.ts
erdirowlands: Thanks! Done.

src/sdk_codes.ts
erdirowlands: Thanks - I've changed the String to:

'Polling stopped'

I found it was difficult to get a handle on the reason, so opted not to do it, but forgot to update the string. 

if it's due to streaming started or polling encountering an error, then we'll get the right codes for those subsequently

src/sdk_codes.ts
erdirowlands: Changed to Authentication attempt failed

src/sdk_codes.ts
erdirowlands: Agreed, the whole string will now be 

`Authentication attempt failed: attempt=3, error=something, continue retrying=true`

erdirowlands: Screenshot 2023-06-23 at 12 25 27

src/sdk_codes.ts
erdirowlands: Done!

src/sdk_codes.ts
erdirowlands: Whoops, forgot to update the string to remove the colon as I opted not to log the event - it's too noisy. This code will only be logged in a `data` event on the SSE stream, the actual event is then logged in due course.

src/evaluator.ts
erdirowlands: In my real testing, I confirmed it was an `Object`
Screenshot 2023-06-23 at 12 42 53
t`

src/metrics.ts
erdirowlands: It's a good point. We haven't done that with Python or Ruby. We probably should. It will require a bit of time which I don't have right now - if it's ok with you, I'll opt to leave the old ones in, and get them removed when I next have time.

src/evaluator.ts
knagurski: If that's the case then line 463 should explode when you try to `JSON.parse` an object
Screenshot 2023-06-23 at 12 49 37

erdirowlands: Blooming heck you're right it is a JSON string already. Sorry. Fixed

src/client.ts
```diff
@@ -372,6 +372,15 @@ export default class Client {
...
     defaultValue = '',
   ): Promise<string> {
+    if (!this.initialized) {
+      warnDefaultVariationServed(
+        identifier,
+        target,
+        defaultValue.toString(),
+        this.log,
+      );
+      Promise.resolve(defaultValue);
```
erdirowlands, line 382: No point trying to evaluate if we haven't finished initializing, so just return default value immediately.

erdirowlands: I have ran further manual tests with `waitForinit` called and not called.

src/client.ts
```diff
@@ -357,6 +424,15 @@ export default class Client {
...
     defaultValue = {},
   ): Promise<Record<string, unknown>> {
+    if (!this.initialized) {
+      warnDefaultVariationServed(
+        identifier,
+        target,
+        defaultValue.toString(),
+        this.log,
+      );
+      Promise.resolve(defaultValue);
```
knagurski, line 434: I think you'll need to return this. As-is this `Promise` looks to be resolving into the void and the `evaluator.jsonVariation` will still get called.

knagurski: Same with the other ones above too

knagurski: Oh and super minor thing, but worth doing is to make sure conditions and returns have whitespace around them to make sure the file can be easily visually scanned. A wall of code makes it difficult to spot things.

erdirowlands: Done on both counts

knagurski (approved):
Looks the part

PR #75, opened 2023-06-26
Maint - remove stale GitHub Action

erdirowlands:
# What
Removes redundant GitHub action 

# Why
We use Harness for pipelines and this is not required.


knagurski (approved):
Kill them with fire 🔥

PR #81, opened 2023-08-07
FFM-8886 Add retries to polling and streaming for fetching flags/groups

erdirowlands:
# What
Adds retries to streaming and polling when fetching flags/groups

Adds retries to metrics requests

# Why
Streaming - if the call to /flags timed out, then the SDK would crash. This adds retry logic, and does not throw an exception which causes the SDK to crash.

Polling - if the same call failed, we'd begin a new polling window. In environments like lambda, this can result in a long wait between flag updates, so this change retries first instead of starting the polling window again.

# Testing
WIP - have tested that streaming now works in lambda correctly. That is to say, a call to a warm instance does not result in an immediate timeout, but rather the call retries and gets the updated flag value.

Still to test polling


src/streaming.ts
```diff
@@ -177,21 +177,47 @@ export class StreamProcessor {
...
+          this.getSegmentByIdentifierWithRetries(3),
           this.repository.setSegment.bind(this.repository),
           this.repository.deleteSegment.bind(this.repository),
         );
       }
     }
   }
 
+  private getFeatureConfigByIdentifierWithRetries(maxRetries: number) {
+    for (let i = 0; i < maxRetries; i++) {
```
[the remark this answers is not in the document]
erdirowlands: It's not. I haven't added any backoff behaviour to the retry. I wasn't sure if it was necessasry for lambda users, because as soon as the instance unfreezes it'll get a timeout error immediately and the first retry will connect as normal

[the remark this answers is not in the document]
erdirowlands: I _ could_ add a backoff to each retry if desired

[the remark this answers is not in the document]
andybharness: i was just thinking about the non-serverless use cases - is it as simple as adding sleep() or with that block the event loop?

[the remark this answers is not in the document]
erdirowlands: Will see what the idomatic approach is for node now

src/streaming.ts
```diff
@@ -195,33 +198,51 @@ export class StreamProcessor {
...
         delFn(msg.identifier);
       }
     } catch (error) {
       this.log.error(
-        'Error while fetching data with identifier:',
+        `Error while fetching ${fetchType} with identifier:`,
         msg.identifier,
         error,
       );
-      throw error;
```
erdirowlands, line 219: We shouldn't have been doing this, as unless the Client was wrapped in the user's own try-catch logic, this would have crashed the customer's app

src/streaming.ts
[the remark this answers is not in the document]
erdirowlands: Yeah the approach is to do the following:

await new Promise(res => setTimeout(res, delay));

That won't block the event loop. It can, however, cause a resource contention issue if the function is doing loads of retries -i.e. in a lambda environment. So I'd opt to leave it out just in case. 
The function will retry max 5 times, and if it fails, we won't try anymore.

src/streaming.ts
```diff
@@ -143,7 +143,6 @@ export class StreamProcessor {
 
         res
           .on('data', (data) => {
-            debugStreamEventReceived(this.log);
```
erdirowlands, line 146: Stops heartbeats from being logged

src/metrics.ts
```diff
@@ -182,33 +182,53 @@ export const MetricsProcessor = (
...
   };
 
+  async function postMetricsWithRetry(
+    environment: string,
+    cluster: string,
+    metrics: Metrics,
+    retries = 3,
+  ) {
+    // specify the return type explicitly
+    let lastError: Error | null = null;
```
erdirowlands, line 214:
```suggestion
    let lastError: Error | null = null;
```

src/metrics.ts
```diff
@@ -182,33 +182,52 @@ export const MetricsProcessor = (
...
+    retries = 3,
+  ) {
+    let lastError: Error | null = null;
+    for (let attempt = 1; attempt <= retries; attempt++) {
+      try {
+        const response = await api.postMetrics(environment, cluster, metrics);
+        if (response.status >= 400 && response.status <= 599) {
+          throw new Error(`HTTP Error: ${response.status}`);
+        }
+        return response; // Exit function if successful.
```
erdirowlands, line 220:
```suggestion
        return response; 
```

src/client.ts
```diff
@@ -71,6 +70,7 @@ export default class Client {
     this.sdkKey = sdkKey;
     this.options = { ...defaultOptions, ...options };
     this.log = this.options.logger;
+    axios.defaults.timeout = this.options.axiosTimeout;
```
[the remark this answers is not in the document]
erdirowlands: I'll need to update the README and developer hub docs with this new config option

src/metrics.ts
```diff
@@ -182,33 +182,53 @@ export const MetricsProcessor = (
     };
   };
 
-  const _send = (): void => {
+  const _send = async () => {
```
knagurski, line 185: Might as well keep a return type
```suggestion
  const _send = async (): Promise<void> => {
```

src/constants.ts
```diff
@@ -39,6 +39,7 @@ export const defaultOptions: Options = {
   eventsUrl: EVENTS_URL,
   pollInterval: PULL_INTERVAL,
   eventsSyncInterval: EVENTS_SYNC_INTERVAL,
+  axiosTimeout: 30000,
```
knagurski, line 42: Can we extract this to a declared constant like the other intervals?

src/metrics.ts
```diff
@@ -182,33 +182,53 @@ export const MetricsProcessor = (
...
-              response.status,
-            );
-          }
-        })
-        .catch((error: Error) => {
-          warnPostMetricsFailed(`${error}`, log);
-          log.debug('Metrics server returns error: ', error);
-        });
+      try {
+        const response = await postMetricsWithRetry(
```
knagurski, line 194: Just out of curiosity, why change from using the Promise methods to async/await?

src/metrics.ts
```diff
@@ -182,33 +182,53 @@ export const MetricsProcessor = (
...
+          environment,
+          cluster,
+          metrics,
+        );
+        log.debug('Metrics server returns:', response.status);
+        infoMetricsSuccess(log);
+      } catch (error) {
+        warnPostMetricsFailed(`${error}`, log);
```
knagurski, line 202: Does this need to be a template string, or is it just being used as a quick cast to string?

src/metrics.ts
```diff
@@ -182,33 +182,53 @@ export const MetricsProcessor = (
...
+        return response; // Exit function if successful.
+      } catch (error) {
+        lastError = error;
+        if (attempt === retries) {
+          log.debug('Failed to send metrics after retries:', error.message);
+        }
+      }
+    }
+    throw lastError;
+  }
```
knagurski: This is a bit of a wall of text. Can we add some line breaks make it more readable?
```suggestion
  async function postMetricsWithRetry(
    environment: string,
    cluster: string,
    metrics: Metrics,
    retries = 3,
  ) {
    // specify the return type explicitly
    let lastError: Error | null = null;
    
    for (let attempt = 1; attempt <= retries; attempt++) {
      try {
        const response = await api.postMetrics(environment, cluster, metrics);
        if (response.status >= 400 && response.status <= 599) {
          throw new Error(`HTTP Error: ${response.status}`);
        }
        
        return response; // Exit function if successful.
      } catch (error) {
        lastError = error;
        
        if (attempt === retries) {
          log.debug('Failed to send metrics after retries:', error.message);
        }
      }
    }
    
    throw lastError;
  }
```

src/polling.ts
```diff
@@ -90,41 +90,60 @@ export class PollingProcessor {
...
-        this.cluster,
-      );
-      this.log.debug('Fetching flags finished');
-      response.data.forEach((fc: FeatureConfig) =>
-        this.repository.setFlag(fc.feature, fc),
-      );
-    } catch (error) {
-      this.log.error('Error loading flags', error);
-      throw error;
+  private async retrieveFlags(retries = 3): Promise<void> {
```
knagurski, line 93: This function is a bit of a wall of text too and difficult to visually scan.

src/polling.ts
```diff
@@ -90,41 +90,60 @@ export class PollingProcessor {
...
+        this.log.debug('Fetching flags');
+        const response = await this.api.getFeatureConfig(
+          this.environment,
+          this.cluster,
+        );
+        this.log.debug('Fetching flags finished');
+        response.data.forEach((fc: FeatureConfig) =>
+          this.repository.setFlag(fc.feature, fc),
+        );
+        break;
```
knagurski, line 105: Not a problem, but returning from the function might be a better option than breaking the loop. The break could, ummm, break, if another breakable scope wraps this, where as the return is pretty definitive.

src/streaming.ts
```diff
@@ -195,33 +198,51 @@ export class StreamProcessor {
...
         error,
       );
-      throw error;
     }
-    this.log.info('Processing message finished', msg);
+    this.log.info(`Processing message with type ${fetchType} finished`, msg);
     return;
   }
 
+  private async retryFetchOperation(fn, fetchType: string, retries = 5) {
```
knagurski, line 227: Curious why we retry 5 times here, but 3 times elsewhere?

src/polling.ts
knagurski, line 93: `<padantic-alert>`
This will actually only _retry_ twice, but _try_ thrice.
`</padantic-alert>`

Not a problem. But it's worth setting expectations that this will not "retry" 3 times.

src/streaming.ts
```diff
@@ -175,16 +174,20 @@ export class StreamProcessor {
       const msg = JSON.parse(line.substring(5));
 
       if (msg.domain === 'flag') {
+        debugStreamEventReceived(this.log);
```
knagurski, line 177: Curious why you moved this inside the `if`'s? It _might_ be useful to log when messages are not a known `domain`. Not a blocker at all.

src/constants.ts
erdirowlands: Done: https://github.com/harness/ff-nodejs-server-sdk/commit/cb549e33d6acf0d7923bf71ec940134f36477ffd

src/metrics.ts
erdirowlands: I think it was just an after effect of hacking to get it working correctly, and I inadvertantly changed it to async/await. No other reason. But it loos like the only use of a promise method in this file - so at least it didn't break the pattern.

src/metrics.ts
erdirowlands: Yikes, you are right. Done - it wouldn't let me commit your suggestion as I commited your other suggestion which affected this snippet. Here's the commit I made: https://github.com/harness/ff-nodejs-server-sdk/commit/d25dfc413bee4795906707a5cf250c54c03ac49e

src/polling.ts
erdirowlands: Fixed: https://github.com/harness/ff-nodejs-server-sdk/commit/cccd3e8127c3c9e1828b73f6c774974bd111819b

src/polling.ts
erdirowlands: Updated to return in this function and it's segment counterpart

erdirowlands: https://github.com/harness/ff-nodejs-server-sdk/commit/b72eb5040ff7c227d0a7250e1c7df924581ca490

src/polling.ts
erdirowlands: Flip - good spot. What would you change the name to in this sort of situation? Or would you alter the code so that it did retry 3 times?

src/streaming.ts
erdirowlands: Cool - it was to get rid of that pesky heartbeat log. But, I've updated it so we only log messages if they have a `domain` property, e.g. flags/segments/others but not heartbeats. https://github.com/harness/ff-nodejs-server-sdk/commit/f8a2c99878ba56207f8ec2c447014981552fb803

src/client.ts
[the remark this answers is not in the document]
erdirowlands: Updated the README. Developer hub will be tricky as we have lost Gigi. if it's ok, will just go readme for now.

src/polling.ts
knagurski: It's totally a non-issue. But if you want to remove any doubt, you can drop the `re` part and keep it as `tries`.

erdirowlands: I should have probably changed `postMetricsWithRetries` to `postMetricsWithMultipleAttempts` too... but that just sounds bad. THought the original was ok.

erdirowlands:
I'm going to run a full round of testing the retry behaviour of polling/streaming/metrics against a local proxy which simulates failures, before merging this.

erdirowlands:
Closing PR as stale. Will revisit.

PR #83, opened 2023-10-01
fix: nodenext type export

### LINKED ISSUE #80: NodeNext support
`exports.types` is missing in package.json
So when typescript `moduleResolution` is set to `nodenext`, types are not able to recognize by typescripts

marklai1998:
close #80


erdirowlands (approved):
This looks good. Thank you for raising the issue, and for the PR.

PR #84, opened 2023-10-01
fix: add type as deps

marklai1998:
image

Because `lru-cache` is referenced in `ff-nodejs-server-sdk` definition file and "@types/lru-cache" is not deps

`tsc` will fail without setting `skipLibChecks` to true on the user side

or installing both `@types/lru-cache` and `lru-cache` on the user side

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
       ],
       "license": "Apache-2.0",
       "dependencies": {
+        "@types/lru-cache": "^7.10.10",
         "axios": "^0.22.0",
         "axios-retry": "^3.2.0",
         "jwt-decode": "^3.1.2",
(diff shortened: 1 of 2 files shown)


erdirowlands (approved):
Looks good - we currently export this like you say, so happy to include the type defs for it.

marklai1998:
> Looks good - we currently export this like you say, so happy to include the type defs for it.

Thanks, when are these fixes gonna merge and deploy?

erdirowlands:
> > Looks good - we currently export this like you say, so happy to include the type defs for it.
> 
> Thanks, when are these fixes gonna merge and deployed

Illget these 

> > Looks good - we currently export this like you say, so happy to include the type defs for it.
> 
> Thanks, when are these fixes gonna merge and deploy?

I'll get both fixes merged and a new SDK version released to NPM tomorrow morning UK time.

PR #87, opened 2023-10-05
fix: remove lru cache type

marklai1998:
Follow up on https://github.com/harness/ff-nodejs-server-sdk/pull/84

The issue still happening, after further inspection of the @types it installed, the type is completely empty

image

After going to the npm page, the package is deprecated, lru-cache has moved to ts.
https://www.npmjs.com/package/@types/lru-cache

So by removing the @types it will work as expected

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
       ],
       "license": "Apache-2.0",
       "dependencies": {
-        "@types/lru-cache": "^7.10.10",
         "axios": "^0.22.0",
         "axios-retry": "^3.2.0",
         "jwt-decode": "^3.1.2",
@@ -1768,15 +1767,6 @@
       "license": "MIT",
       "dependencies": {
         "@types/node": "*"
-      }
-    },
-    "node_modules/@types/lru-cache": {
-      "version": "7.10.10",
-      "resolved": "https://registry.npmjs.org/@types/lru-cache/-/lru-cache-7.10.10.tgz",
-      "integrity": "sha512-nEpVRPWW9EBmx2SCfNn3ClYxPL7IktPX12HhIoSc/H5mMjdeW3+YsXIpseLQ2xF35+OcpwKQbEUw5VtqE4PDNA==",
-      "deprecated": "This is a stub types definition. lru-cache provides its own type definitions, so you do not need this installed.",
-      "dependencies": {
-        "lru-cache": "*"
       }
     },
     "node_modules/@types/node": {
@@ -8209,14 +8199,6 @@
       "dev": true,
       "requires": {
         "@types/node": "*"
-      }
-    },
-    "@types/lru-cache": {
-      "version": "7.10.10",
-      "resolved": "https://registry.npmjs.org/@types/lru-cache/-/lru-cache-7.10.10.tgz",
-      "integrity": "sha512-nEpVRPWW9EBmx2SCfNn3ClYxPL7IktPX12HhIoSc/H5mMjdeW3+YsXIpseLQ2xF35+OcpwKQbEUw5VtqE4PDNA==",
-      "requires": {
-        "lru-cache": "*"
       }
     },
     "@types/node": {
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "url": "https://github.com/harness/ff-nodejs-server-sdk"
   },
   "dependencies": {
-    "@types/lru-cache": "^7.10.10",
     "axios": "^0.22.0",
     "axios-retry": "^3.2.0",
     "jwt-decode": "^3.1.2",
--- a/package.json
+++ b/package.json
@@ -85,5 +85,11 @@
   ],
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "bundleDependencies": [
+    "axios",
+    "eventsource",
+    "keyv",
+    "keyv-file"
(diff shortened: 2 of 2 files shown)


erdirowlands:
Hi @marklai1998  

I see where you're coming from. We noticed warnings that the `lru` package provided its own types, but were willing to provide them as we currently export it. However, that clearly hasn't worked for you for the reason you mentioned.

We're keen to get this resolved for you. I've went ahead and released `1.3.4-rc.1` which incoporates the same fix as you've made here. Can you give that a try and let me know if it works first, before I merge this PR?

marklai1998:
> Hi @marklai1998
> 
> I see where you're coming from. We noticed warnings that the `lru-cache` package provided its own types, but were willing to provide them as we currently export it. However, that clearly hasn't worked for you for the reason you mentioned.
> 
> We're keen to get this resolved for you. I've went ahead and released `1.3.4-rc.1` which incoporates the same fix as you've made here. Can you give that a try and let me know if it works first, before I merge this PR?

Good to point out, cuz lru-cache is on 10.x now where the installed version here is 6.x that haven migrated to ts yet
so we should consider upgrading lru-cache also

marklai1998:
@erdirowlands can you take some time to upgrade it? There are a few breaking version that need code change here
I'm not that familiar with the code base

erdirowlands:
> @erdirowlands can you take some time to upgrade it? There are a few breaking version that need code change here I'm not that familiar with the code base

I believe version 8 has been re-written in TS, which requires Node v16 and above. We currently support Node v12 in our SDK, so upgrading to version 8 would require some investigation so we don't impact some users.

Just checking, did the release candidate version work for you?

marklai1998:
> > @erdirowlands can you take some time to upgrade it? There are a few breaking version that need code change here I'm not that familiar with the code base
> 
> I believe version 8 has been re-written in TS, which requires Node v16 and above. We currently support Node v12 in our SDK, so upgrading to version 8 would require some investigation so we don't impact some users.
> 
> Just checking, did the release candidate version work for you?

I see, then I think we should downgrade the @type dependency to a workable version

Will check out the release candidate soon, but in theory, it won't work because of the V6 dependency

marklai1998:
> > @erdirowlands can you take some time to upgrade it? There are a few breaking version that need code change here I'm not that familiar with the code base
> 
> I believe version 8 has been re-written in TS, which requires Node v16 and above. We currently support Node v12 in our SDK, so upgrading to version 8 would require some investigation so we don't impact some users.
> 
> Just checking, did the release candidate version work for you?

just tried it, indeed without `lru-cache` installed on the client side it will throw error since the deps is V6 with no type
but with `lru-cache` V10 installed on the client there are no problem

The downgrade @type method should be the way to go

erdirowlands:
> > > @erdirowlands can you take some time to upgrade it? There are a few breaking version that need code change here I'm not that familiar with the code base
> > 
> > 
> > I believe version 8 has been re-written in TS, which requires Node v16 and above. We currently support Node v12 in our SDK, so upgrading to version 8 would require some investigation so we don't impact some users.
> > Just checking, did the release candidate version work for you?
> 
> just tried it, indeed without `lru-cache` installed on the client side it will throw error since the deps is V6 with no type but with `lru-cache` V10 installed on the client there are no problem
> 
> The downgrade @type method should be the way to go

I've released `1.3.4-rc.2` which uses `lru-cache` 7.0.0 - this version comes bundled with its own types.

Can you give this a go and let me know if it solves the problem. 

if so, I'll get this moved through our internal code review as it required refactoring some of the SDK code. I'll aim to have a new release out late next week.

marklai1998:
> 1.3.4-rc.2

the `keys` function has changed to returning Generator<string, void, void> on lru-cache 7.0.0, which has conflicts with ff-client "AsyncKeyValueStore" type

erdirowlands:
> > 1.3.4-rc.2
> 
> the `keys` function has changed to returning Generator<string, void, void> on lru-cache 7.0.0, which has conflicts with ff-client "AsyncKeyValueStore" type

Hmm, I updated the signature. I must have not built the SDK before releasing. Sorry. Let me publish a new version.

erdirowlands:
Sorry about that @marklai1998 

Can you please try 1.3.4-rc.3

marklai1998:
> 1.3.4-rc.3

Sorry, I didn't read the error well, is the `set` call signature incompatible not `keys()`

erdirowlands:
> > 1.3.4-rc.3
> 
> Sorry, I didn't read the error well, is the `set` call signature incompatible not `keys()`

What's the exact error you're seeing? I don't have any issues with `tsc` here.

marklai1998:
> > > 1.3.4-rc.3
> > 
> > 
> > Sorry, I didn't read the error well, is the `set` call signature incompatible not `keys()`
> 
> What's the exact error you're seeing? I don't have any issues with `tsc` here.

image

AsyncKeyValueStore expect set() return Promise<true> while LRU return `this`

erdirowlands:
Thanks. You must be operating in a much stricter typescript environmental than us.

I'll look at fixing this on Monday. 

Could you also share your ts.config file via the support ticket you raised, so that I can attempt to reproduce any typescript errors. 

I'll send an update via the support ticket. For now,.I will close this PR.

Thanks for your patience.

PR #88, opened 2023-10-06
FFM-9574 Upgrade `lru-cache` to version 7.0.0

erdirowlands:
# What
Upgrades `lru-cache` library to 7.0.0 

# Why
We export the `lru-cache` and a user could not `tsc` their project with our SDK installed. This is a valid issue, because version 6 of the library did not come with type defentions, and we were only providing the types as dev deps. 

We tried providing them as dependencies, but there was a mismatch bettwen the type defenition version and the library version. 

This version includes its own types, and comes with a big performance upgrade, but it required a minor refactor of `keys` and `del` 

# Testing
Manual - streaming, polling. Will carry out more extensive testing like deleting/updating flags and target groups etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.3",
+  "version": "1.3.4-rc.1",
   "description": "Feature flags SDK for NodeJS environments",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
@@ -61,7 +61,6 @@
     "url": "https://github.com/harness/ff-nodejs-server-sdk"
   },
   "dependencies": {
-    "@types/lru-cache": "^7.10.10",
     "axios": "^0.22.0",
     "axios-retry": "^3.2.0",
     "jwt-decode": "^3.1.2",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "jwt-decode": "^3.1.2",
     "keyv": "^4.0.3",
     "keyv-file": "^0.2.0",
-    "lru-cache": "^6.0.0",
+    "lru-cache": "^7.0.0",
     "murmurhash": "^2.0.0",
     "tslib": "~2.3.0"
   },
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -11,11 +11,13 @@ export class SimpleCache implements KeyValueStore {
     return this.cache[key];
   }
 
-  del(key: string): void {
+  delete(key: string): void {
     delete this.cache[key];
   }
 
-  keys(): string[] {
-    return Object.keys(this.cache);
+  *keys(): Generator<string, void, void> {
+    for (const key of Object.keys(this.cache)) {
+      yield key;
+    }
   }
 }
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,1 +1,1 @@
-export const VERSION = '1.3.3';
+export const VERSION = "1.3.4-rc.1";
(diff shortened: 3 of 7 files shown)


src/cache.ts
```diff
@@ -11,11 +11,13 @@ export class SimpleCache implements KeyValueStore {
     return this.cache[key];
   }
 
-  del(key: string): void {
+  delete(key: string): void {
```
erdirowlands, line 14: `del` is deprecated in version 7

package.json
```diff
@@ -61,13 +61,12 @@
...
   },
   "dependencies": {
-    "@types/lru-cache": "^7.10.10",
     "axios": "^0.22.0",
     "axios-retry": "^3.2.0",
     "jwt-decode": "^3.1.2",
     "keyv": "^4.0.3",
     "keyv-file": "^0.2.0",
-    "lru-cache": "^6.0.0",
+    "lru-cache": "^7.0.0",
```
erdirowlands, line 69: we could probably go to the highest minor version for 7 here (8 required node v16). I'll update and test on Monday

src/version.ts
```diff
@@ -1 +1 @@
-export const VERSION = '1.3.3';
+export const VERSION = "1.3.4-rc.1";
```
erdirowlands, line 1: Accidentally checked in the RC version. Will remove this pre-merge

examples/getting_started/index.js
knagurski, line 1: Did you mean to delete these files?

erdirowlands: No 🤦

src/index.ts
```diff
@@ -1,5 +1,5 @@
 import Client, { Event } from './client';
-import LRU from 'lru-cache';
+import * as LRU from 'lru-cache';
```
erdirowlands, line 2: The library uses both CommonJS and ESM exports, and the customer who raised the issue does not allow ESM imports explicitly in their typescript config, so I've used this syntax.

knagurski: That's weird. Typically people go the other way since TS _only_ allows the ESM format (though it can resolve both)

andybharness (approved):

src/cache.ts
```diff
@@ -11,11 +11,13 @@ export class SimpleCache implements KeyValueStore {
...
     delete this.cache[key];
   }
 
-  keys(): string[] {
-    return Object.keys(this.cache);
+  *keys(): Generator<string, void, void> {
+    for (const key of Object.keys(this.cache)) {
+      yield key;
+    }
```
knagurski, line 21: Ohhh, haven't seen any one use generators in JS for a while. A forgotten feature 👍🏻

src/index.ts
erdirowlands: Very weird... 
I used their ts.config file and ran `tsc myfile.ts` and it complained about that. With   `"esModuleInterop": true` it was fine

knagurski: Modules in Node are in a terrible place at the moment. That's where it's losing a lot of devs to things like Deno and Bun since they do a _much_ better job outa the box. ¯\_(ツ)_/¯. This is good though, I'd have imported as `* as LRU` anyway unless the package specifically called out that the default export was an object. Destructuring in an `import` _looks_ like normal destructuring, but it's actually very different.

erdirowlands:
I've carried out some further testing on top of what I've said above:

1. Add and remove target from a flag. 
2. Add and remove target from a group.
3. Delete target group - there is an issue here which @davejohnston discovered that we need to put a fix in for, where if you delete a target group, the event doesn't get propegated up correctly. I'm hardening Node after I finish Golan, so will fix this then.

erdirowlands:
I've released `1.3.4-rc.4` and currently waiting for the customer to confirm everything compiles correctly before merging and releasing this.

erdirowlands:
Customer confirms issues resolved. Merging.

PR #89, opened 2023-10-16
[FFM-9621] - Avoid sending empty metric payloads

andybharness:
[FFM-9621] - Avoid sending empty metric payloads

What
Adds a check to make sure metrics data is not empty before sending it.

Why
We're seeing empty metrics payloads arriving at the BE we should avoid wasting network bandwidth.

Testing
Manual

[FFM-9621]: https://harness.atlassian.net/browse/FFM-9621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -189,7 +189,7 @@ export const MetricsProcessor = (
 
     const metrics: Metrics = _summarize();
 
-    if (metrics && metrics.metricsData.length + metrics.targetData.length > 0) {
+    if (metrics && metrics.metricsData.length && metrics.targetData.length) {
       log.debug('Start sending metrics data');
       api
         .postMetrics(environment, cluster, metrics)
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -188,7 +188,8 @@ export const MetricsProcessor = (
     }
 
     const metrics: Metrics = _summarize();
-    if (metrics) {
+
+    if (metrics && metrics.metricsData.length + metrics.targetData.length > 0) {
       log.debug('Start sending metrics data');
       api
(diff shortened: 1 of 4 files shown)


ribeirophillipe (approved):

src/metrics.ts
```diff
@@ -188,7 +188,8 @@ export const MetricsProcessor = (
     }
 
     const metrics: Metrics = _summarize();
-    if (metrics) {
+
+    if (metrics && metrics.metricsData.length + metrics.targetData.length > 0) {
```
knagurski, line 192: Super minor and works the same way, but you can treat these as `truthy`/`falsey`.
```suggestion
    if (metrics && metrics.metricsData.length && metrics.targetData.length) {
```

knagurski: That is unless you do want to send if `metricsData` is empty, but `targetData` isn't. That that case, your could `||` them
```suggestion
    if (metrics && (metrics.metricsData.length || metrics.targetData.length)) {
```

knagurski:
Do you need to bump the version as well for a release?

andybharness:
I've only bumped the minor since there's no real API changes - let me double check if those files were missed

andybharness:
@knagurski i had been working on a older copy of the git tree and rev'ed the version to 1.3.4, turns out that version had been released recently. I've fixed the version now to be 1.3.5

knagurski (approved):
Nice and simple 👍🏻

PR #90, opened 2023-10-30
FFM-9761 Stringify Auth Error

erdirowlands:
# What
For auth errors, stringify the error so that it appears on the same line. 

# Why
In cloud logging, the `error` object appears on multiple lines. 

# Testing
Will release an RC for the internal team to check

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -233,7 +233,7 @@ export default class Client {
       this.cluster = decoded.clusterIdentifier || '1';
     } catch (error) {
       this.failure = true;
-      console.error('Error while authenticating, err: ', error);
+      console.error('Error while authenticating, err: ' + JSON.stringify(error));
       warnAuthFailedSrvDefaults(this.log);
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -233,7 +233,7 @@ export default class Client {
       this.cluster = decoded.clusterIdentifier || '1';
     } catch (error) {
       this.failure = true;
-      console.error('Error while authenticating, err: ', error);
+      console.error('Error while authenticating, err: ' + JSON.stringify(error));
```
knagurski, line 236: Passing it as its original value should means it's inspectable in a console as well as serialised in logs. The failure of the display in the GCP console is more of a problem with their logging, not an issue with our SDK. I'd move this from a bug ticket to a chore or story and bump the version as a minor rather than a patch. This isn't a bug

erdirowlands:
Closing, see JIRA ticket.

PR #91, opened 2023-10-30
FFM-9778 Update some log statements from console to the SDK's logger

erdirowlands:
# What
Two log statements were using `console` instead of the `Logger` interface, this PR updates those references. 

# Why
So any customer loggers that are provides work across all log statements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.5",
+  "version": "1.3.6",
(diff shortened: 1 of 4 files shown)


andybharness (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.5",
+  "version": "1.3.6",
```
knagurski, line 3: Since this wasn't a bug, it shouldn't have been versioned as one. Ah well

PR #99, opened 2023-11-23
[FFM-10041] - Fall back to identifier if bucketBy attr is not found

andybharness:
[FFM-10041] - Fall back to identifier if bucketBy attribute is not found

**What**
Update BucketBy logic to fall back to the target identifier if given BucketBy attribute is not found.
Adds new unit test to test fallback logic works.

**Why**
Keep the SDK consistent with the behaviour of the other server SDKs.

**Testing**
Testgrid + new unit test

[FFM-10041]: https://harness.atlassian.net/browse/FFM-10041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.8",
+  "version": "1.4.0",
   "description": "Feature flags SDK for NodeJS environments",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.7",
+  "version": "1.3.8",
   "description": "Feature flags SDK for NodeJS environments",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
--- a/src/__tests__/evaluator.test.ts
+++ b/src/__tests__/evaluator.test.ts
@@ -1,0 +1,80 @@
+import { Evaluator } from '../evaluator';
+import { Logger } from '../log';
+
+describe('Evaluator', () => {
+  it('should return identifier if bucketby attribute does not exist', async () => {
+    const logger: Logger = {
+      trace: jest.fn(),
+      debug: jest.fn(),
+      info: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+    };
+
+    const mockQuery = {
+      getFlag: jest.fn(),
+      getSegment: jest.fn(),
+      findFlagsBySegment: jest.fn(),
+    };
+
+    const featureConfig = {
+      feature: 'flag',
+      kind: 'string',
+      state: 'on',
+      variationToTargetMap: [],
+      variations: [
+        { identifier: 'variation1', value: 'default_on' },
+        { identifier: 'variation2', value: 'wanted_value' },
+        { identifier: 'variation3', value: 'default_off' },
+      ],
+      defaultServe: {
+        distribution: null,
+        variation: 'default_on',
+      },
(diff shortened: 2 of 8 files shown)


erdirowlands (approved):

src/__tests__/evaluator.test.ts
knagurski, line 1: Awesome job making sure tests are included!

src/evaluator.ts
```diff
@@ -76,9 +80,19 @@ export class Evaluator {
     bucketBy: string,
     percentage: number,
   ): boolean {
-    const property = this.getAttrValue(target, bucketBy);
+    let property = this.getAttrValue(target, bucketBy);
     if (!property) {
-      return false;
+      const old_bucketBy = bucketBy;
```
knagurski, line 85: The standard in JS is camel-case, not snake
```suggestion
      const oldBucketBy = bucketBy;
```

src/evaluator.ts
```diff
@@ -76,9 +80,19 @@ export class Evaluator {
...
+      bucketBy = 'identifier';
```
knagurski, line 86: There's technically no reason you can't do this as args are considered to be `var` (as opposed to `let` or `const`), but it's generally a bad idea to mutate args. Maybe better to extract a `let` variable as you did with `property`

src/evaluator.ts
```diff
@@ -76,9 +80,19 @@ export class Evaluator {
...
+      property = this.getAttrValue(target, bucketBy);
+      if (!property) {
+        return false;
+      }
+      warnBucketByAttributeNotFound(
```
knagurski, line 91: This is becoming a bit of a wall of code. A few line breaks around the `if` will make it a lot easier to visually scan.
```suggestion
      property = this.getAttrValue(target, bucketBy);
      
      if (!property) {
        return false;
      }
      
      warnBucketByAttributeNotFound(
```

src/evaluator.ts
```diff
@@ -76,9 +80,19 @@ export class Evaluator {
...
+        old_bucketBy,
+        property?.toString(),
+        this.log,
+      );
     }
     const bucketId = this.getNormalizedNumber(property, bucketBy);
```
knagurski, line 97: If I remember correctly, Node has the property and bucket by backwards in this, can we verify and fix that at the same time?

src/__tests__/evaluator.test.ts
```diff
@@ -0,0 +1,80 @@
+import { Evaluator } from '../evaluator';
+import { Logger } from '../log';
+
+describe('Evaluator', () => {
+  it('should return identifier if bucketby attribute does not exist', async () => {
```
knagurski, line 5: Technically this does not return identifier if bucketBy attribute doesn't exist, buckets by the identifier if the bucketBy attribute doesn't exist and returns the appropriate variation value.

src/evaluator.ts
andybharness: lol - I work on so many languages these days I've no idea anymore

src/evaluator.ts
andybharness: do you mean to keep parameter ordering consistent consistent with other SDKs? yeah should be easy enough

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harnessio/ff-nodejs-server-sdk",
-  "version": "1.3.7",
+  "version": "1.3.8",
```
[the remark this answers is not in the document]
andybharness: Makes sense - I've been bumping the second digit for other SDKs but only if the hash output changed, NodeJS had the correct hash but needed this fallback code. Should be easy to update

knagurski (approved):
👍🏻

PR #100, opened 2024-01-03
Fix metric not sending when metric target is empty

weisheng-homage:
An issue arises if all your feature flag query calls do not have target, this will result in empty targetData, but with data in metricsData.

This fix metric processor so that it would send metric, if either metricsData or targetData has something in it.

This is raised in a zendesk support ticket 55786.

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -189,7 +189,7 @@ export const MetricsProcessor = (
 
     const metrics: Metrics = _summarize();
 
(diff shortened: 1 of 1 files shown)


erdirowlands:
Thanks for submitting this PR. 

Per my comment in the support ticket, I will close this, and work on a fix internally. 

Please update the support ticket with any other questions.

PR #101, opened 2024-01-03
FFM-10325 Fix issue with detecting empty evaluation/target metrics

erdirowlands:
# What
Tidies up logic around checking if there are valid metrics to process. This also ensures that metrics are processed and sent when a target hasn't been used for an evaluation. 

Also fixes an issue where we log `5002` on SSE heartbeat events

# Why
A customer reported that when evaluating without a target, their flag metrics were not being reported. 

# Testing 
- TestGrid metrics suite
- Manual sample script:
    -  Evaluation with target
    -  Evaluation with `undefined` target
    -  No evaluations

- SSE Testing:
    - Heartbeats are logged separately
    - Regular events are processed and logged correctly


src/metrics.ts
```diff
@@ -101,11 +101,6 @@ export const MetricsProcessor = (
   };
 
   const _summarize = (): Metrics | unknown => {
-    if (!data) {
```
erdirowlands, line 104: `data` is initialized when it is declared so this will always be true

src/metrics.ts
```diff
@@ -184,30 +179,34 @@ export const MetricsProcessor = (
...
+    }
+
+    if (data.size == 0) {
+      log.debug('No metrics to send in this interval');
       return;
     }
 
     const metrics: Metrics = _summarize();
 
-    if (metrics && metrics.metricsData.length && metrics.targetData.length) {
```
erdirowlands, line 192: We can safely remove this check, because the earlier `data` check now works correctly.

andybharness (approved):

PR #102, opened 2024-01-04
feat: [FFM-10358]: Update and unbundle dependencies

knagurski:
This PR updates several dependencies and unbundles all dependencies. This PR also updates the target version of ECMAscript to match what was available in Node 12.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2019",
     "module": "commonjs",
     "downlevelIteration": true,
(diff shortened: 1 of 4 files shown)


erdirowlands (approved):
Nice 👍

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2019",
```
knagurski, line 3: I thought Node 12 was a lot older than 2019, but it checks out ¯\_(ツ)_/¯. No point in using ES5 when we can go _a lot_ more recent.

PR #106, opened 2024-03-19
FFM-10963 Metrics fix: cast target attribute values to string

erdirowlands:
# What
The Metrics API requires target attribute keys and values to be sent as string.  Previously, the SDK would send attribute values in their original type. This change casts primitives to Strings, and casts an Object to a JSON String.  Note: this is only for metrics only, and the attribute values will appear in the UI in their String format. This format does not affect evaluations in any way. 

Bumps Axios due to CVE with follow-redirects

# Why
To stop `400` errors for invalid metrics requests. 

# Testing
Manual - target with following attributes
`string`
`boolean`
`number`
`bigint`
`symbol`
`array` - all elements comma separated. Mixed typing of elements verified including null.
`object` 
`null` - empty string
`undefined`  - empty string

Axios bump:
Init/streaming/polling/metrics


src/metrics.ts
```diff
@@ -218,6 +225,25 @@ export const MetricsProcessor = (
...
+      case 'string':
+        return value;
+      case 'number':
+      case 'boolean':
+      case 'bigint':
+      case 'symbol':
+        return value.toString();
+      case 'object':
+        if (Array.isArray(value)) {
+          return value.toString()
```
erdirowlands, line 239: For arrays of mixed types, e.g. ["hi", true] this just ensures that the format is in one enclosed string, which is why I didn't use JSON.Stringify.

src/metrics.ts
```diff
@@ -218,6 +225,25 @@ export const MetricsProcessor = (
...
+        }
+        return value === null ? '' : JSON.stringify(value);
+      default:
+        return String(value);
+    }
```
knagurski: I think you can probably simplify this by just leaning on `String` and checking for 
```suggestion
    return typeof value === 'object' && !Array.isArray(value) ? JSON.stringify(value) : String(value)
```

erdirowlands: I very much like this. I tried to simplify it just before your comment. Would you mind refreshing your page and making the suggestion again, and I'll commit it!

erdirowlands: Or I'll just make the change myself

erdirowlands: Done - tested with array, object and !== object. Works well.

knagurski: `null` would throw it off since `null` will return `object` as its type, but you're filtered them out already. Good eye for detail BTW including `Symbol` and `BigInt`, about 99.999% of devs don't even know they exist!

knagurski (approved):
Nice, simple and effective. Great work!

PR #112, opened 2024-04-09
FFM-11022 Applies standard Metrics Enhancements

erdirowlands:
# What
- Splits evaluation and target metrics into their own caches:
    -  so that we can track target metrics correctly. Due to the fact we use the `globalTarget` , this meant that we would miss targets if the evaluation didn't change. 
    - Implements limits: Evaluation metrics to 10K unique evaluations--where a unique evaluation is flag + variation--this allows for evaluation for 10k flags with 5 variations each
    - Target metrics for 100K unique targets
    
- Implement seen targets:
    - ensures that the metrics service only gets sent target metrics for a target once, for the lifetime of the SDK instance
    - ensures that the SDK processes truly unique targets fairly, and that old targets don't occupy the 100K limit set per interval

# Testing
- Metrics accuracy:
    - TestGrid metrics tests
    - Manual sample app
    - Non-anonymous targets appear in UI
    - Anonymous targets don't appear in UI
- SeenTargets:
    - The same target won't be sent on the next interval if it is evaluated against
- Limits:
    - Sent 101K unique targets, only 100K get processed

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -148,15 +148,15 @@ export class MetricsProcessor implements MetricsProcessorInterface {
           key: FEATURE_IDENTIFIER_ATTRIBUTE,
           value: event.featureConfig.feature,
         },
-        { key: FEATURE_NAME_ATTRIBUTE, value: event.featureConfig.feature },
         {
           key: VARIATION_IDENTIFIER_ATTRIBUTE,
           value: event.variation.identifier,
         },
+        { key: FEATURE_NAME_ATTRIBUTE, value: event.featureConfig.feature },
         { key: SDK_TYPE_ATTRIBUTE, value: SDK_TYPE },
         { key: SDK_LANGUAGE_ATTRIBUTE, value: SDK_LANGUAGE },
         { key: SDK_VERSION_ATTRIBUTE, value: VERSION },
-        { key: TARGET_ATTRIBUTE, value: event.target.identifier },
+        { key: TARGET_ATTRIBUTE, value: event?.target?.identifier ?? null },
       ];
 
       const md: MetricsData = {
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -55,7 +55,10 @@ export interface MetricsProcessorInterface {
 
 export class MetricsProcessor implements MetricsProcessorInterface {
   private evaluationAnalytics: Map<string, AnalyticsEvent> = new Map();
-  private targetAnalytics: Map<string, Target> = new Map(); // New cache for Target metrics
+  private targetAnalytics: Map<string, Target> = new Map();
+
+  // Only store and send targets that haven't been sent before in the life of the client instance
+  private seenTargets: Set<string> = new Set();
   private syncInterval?: NodeJS.Timeout;
   private api: MetricsApi;
   private readonly log: Logger;
@@ -120,7 +123,13 @@ export class MetricsProcessor implements MetricsProcessorInterface {
       this.evaluationAnalytics.set(key, event);
     }
 
-    if (target && !target.anonymous) {
+    if (target && target.identifier) {
+      // If target has been seen or is anonymous then ignore it
+      if (this.seenTargets.has(target.identifier) || target.anonymous) {
+        return;
+      }
+
+      this.seenTargets.add(target.identifier);
       this.targetAnalytics.set(target.identifier, target);
     }
   }
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -123,9 +123,10 @@ export class MetricsProcessor implements MetricsProcessorInterface {
       this.evaluationAnalytics.set(key, event);
     }
 
-    if (target && target.identifier) {
-      // If target has been seen or is anonymous then ignore it
-      if (this.seenTargets.has(target.identifier) || target.anonymous) {
+    if (target && !target.anonymous) {
+
+      // If target has been seen then ignore it
+      if (this.seenTargets.has(target.identifier)) {
         return;
       }
 
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -223,10 +223,7 @@ export class MetricsProcessor implements MetricsProcessorInterface {
         );
       }
 
-      let targetName = target.identifier;
-      if (target.name) {
-        targetName = target.name;
-      }
+      const targetName = target.name || target.identifier;
 
       const td: TargetData = {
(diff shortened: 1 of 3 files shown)


src/metrics.ts
```diff
@@ -54,7 +56,18 @@ export interface MetricsProcessorInterface {
...
-  private data: Map<string, AnalyticsEvent> = new Map();
+  private evaluationAnalytics: Map<string, AnalyticsEvent> = new Map();
+  private targetAnalytics: Map<string, Target> = new Map();
+
+  // Only store and send targets that haven't been sent before in the life of the client instance
+  private seenTargets: Set<string> = new Set();
+
+  // Maximum sizes for caches
+  private MAX_EVALUATION_ANALYTICS_SIZE = 10000;
+  private MAX_TARGET_ANALYTICS_SIZE = 100000;
```
knagurski, line 67: I _THINK_ you can set these as `static`. Will make zero difference, but ¯\_(ツ)_/¯

src/metrics.ts
```diff
@@ -107,13 +123,48 @@ export class MetricsProcessor implements MetricsProcessorInterface {
       variation,
       count: 0,
     };
+    this.storeEvaluationAnalytic(event);
+    this.storeTargetAnalytic(target);
+  }
+
+  private storeTargetAnalytic(target: Target) {
```
knagurski, line 130: Totally not an issue, but it's a good practice to add return types
```suggestion
  private storeTargetAnalytic(target: Target): void {
```

src/metrics.ts
```diff
@@ -107,13 +123,48 @@ export class MetricsProcessor implements MetricsProcessorInterface {
...
+      if (this.seenTargets.has(target.identifier)) {
+        return;
+      }
+
+      this.seenTargets.add(target.identifier);
+      this.targetAnalytics.set(target.identifier, target);
+    }
+  }
+
+  private storeEvaluationAnalytic(event: AnalyticsEvent) {
```
knagurski, line 151: Same here
```suggestion
  private storeEvaluationAnalytic(event: AnalyticsEvent): void {
```

andybharness (approved):

src/metrics.ts
```diff
@@ -129,79 +180,62 @@ export class MetricsProcessor implements MetricsProcessorInterface {
...
-        targetData.push(td);
-      }
+    const clonedEvaluationAnalytics = new Map(this.evaluationAnalytics);
+    const clonedTargetAnalytics = new Map(this.targetAnalytics);
+    this.evaluationAnalytics.clear();
+    this.targetAnalytics.clear();
+    this.evaluationAnalyticsExceeded = false;
+    this.targetAnalyticsExceeded = false;
 
+    clonedEvaluationAnalytics.forEach((event) => {
```
knagurski, line 190: TOTALLY not an issue, but instead of creating a variable and then using a `forEach` to push to that variable. You can just swap to `map` and assign in one.
```suggestion
    const metricsData: MetricsData[] = clonedEvaluationAnalytics.map((event) => {
```

src/metrics.ts
```diff
@@ -129,79 +180,62 @@ export class MetricsProcessor implements MetricsProcessorInterface {
...
         timestamp: Date.now(),
         count: event.count,
         metricsType: MetricsDataMetricsTypeEnum.Ffmetrics,
         attributes: metricsAttributes,
       };
       metricsData.push(md);
-    }
+    });
+
+    clonedTargetAnalytics.forEach((target) => {
```
knagurski, line 216: Same here. Also totally a non-issue.
```suggestion
    const targetData: TargetData[] = clonedTargetAnalytics.map((target) => {
```

src/metrics.ts
```diff
@@ -129,79 +180,62 @@ export class MetricsProcessor implements MetricsProcessorInterface {
...
+          ([key, value]) => {
+            return { key, value: this.valueToString(value) };
+          },
+        );
+      }
+
+      let targetName = target.identifier;
+      if (target.name) {
+        targetName = target.name;
+      }
```
knagurski, line 229: Just curious if there was any reason this couldn't be simplified?
```suggestion
      const targetName = target.name || target.identifier;
```

src/metrics.ts
```diff
@@ -129,79 +180,62 @@ export class MetricsProcessor implements MetricsProcessorInterface {
...
+
+      const td: TargetData = {
+        identifier: target.identifier,
+        name: targetName,
+        attributes: targetAttributes,
+      };
+      targetData.push(td);
```
knagurski, line 236: If you do want to keep as a `forEach`/`push`, then you might as well just inline this.
```suggestion
      targetData.push({
        identifier: target.identifier,
        name: targetName,
        attributes: targetAttributes,
      });
```

src/metrics.ts
```diff
@@ -214,7 +248,7 @@ export class MetricsProcessor implements MetricsProcessorInterface {
       return;
     }
 
-    if (this.data.size === 0) {
+    if (this.evaluationAnalytics.size === 0) {
```
knagurski, line 251: Nothing wrong here, but `=== 0` always irks me for some reason.
```suggestion
    if (!this.evaluationAnalytics.size) {
```

knagurski (approved):
Just some comments/suggestions to help with maintainability. No blockers

src/metrics.ts
erdirowlands: I like 🚀  thanks.

erdirowlands: Dang - the `map` function isn't available on `Map` types.  I'd need to convert `clonedEvaluationAnalytics` to an array.  I'll opt to leave it as is.  But in principle I agree with the functional style here.

src/metrics.ts
erdirowlands: Nope - this works better.

src/metrics.ts
erdirowlands: Done

src/metrics.ts
knagurski: Ah man... annoying. You could spread into an array or use `Array.from`. Don't worry about it, it's fine as-is.
```suggestion
    const metricsData: MetricsData[] = [...clonedEvaluationAnalytics.values()].map((event) => {
```

src/metrics.ts
erdirowlands: I researched this - and I agree.

PR #113, opened 2024-04-18
[FFM-11243] - Target v2: Adding support for AND/OR in clauses

andybharness:
[FFM-11243] - Target v2: Adding support for AND/OR in clauses

**What**
Adding support for processing the new `GroupServingRule` in `Segment`. Also update API definition for OpenAPIGenerator plugin. The default behaviour now is to check for the existence of `GroupServingRule` if found then we ignore the old `Rules` section. Old `Rules` section will continue to work to aid transition if `GroupServingRule` is not yet populated. Each entry in `GroupServingRule` is ORed while each list of clauses is ANDed. The `GroupServingRule` with the highest priority used first to allow reordering on the backend.

**Why**
This gives the customer more flexibility on how they configure their rules in the backend.

**Testing**
Initial unit tests for now + testgrid for existing rules

[FFM-11243]: https://harness.atlassian.net/browse/FFM-11243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/__tests__/evaluator.test.ts
+++ b/src/__tests__/evaluator.test.ts
@@ -233,7 +233,6 @@ describe('Evaluator', () => {
       flagName,
       target,
       !expected,
-      null,
     );
 
     expect(result).toEqual(expected);
--- a/src/__tests__/evaluator.test.ts
+++ b/src/__tests__/evaluator.test.ts
@@ -1,16 +1,24 @@
 import { Evaluator } from '../evaluator';
 import { Logger } from '../log';
+import { SimpleCache } from '../cache';
+import { StorageRepository } from '../repository';
+import {
+  FeatureConfig,
+  FeatureConfigKindEnum,
+  FeatureState,
+  Segment,
+} from '../openapi';
+
+const logger: Logger = {
+  trace: jest.fn(),
+  debug: jest.fn(),
+  info: jest.fn(),
+  warn: jest.fn(),
+  error: jest.fn(),
+};
 
 describe('Evaluator', () => {
   it('should fallback to identifier if bucketby attribute does not exist', async () => {
-    const logger: Logger = {
-      trace: jest.fn(),
-      debug: jest.fn(),
-      info: jest.fn(),
-      warn: jest.fn(),
-      error: jest.fn(),
-    };
-
     const mock_query = {
       getFlag: jest.fn(),
       getSegment: jest.fn(),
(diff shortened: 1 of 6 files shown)


src/evaluator.ts
```diff
@@ -152,17 +152,31 @@ export class Evaluator {
...
-        // Should Target be included via segment rules
-        if (
-          segment.rules &&
-          (await this.evaluateClauses(segment.rules, target))
-        ) {
-          this.log.debug(
-            'Target %s included in segment %s via rules\n',
-            target.name,
-            segment.name,
+        if (segment.servingRules && segment.servingRules.length > 0) {
```
erdirowlands, line 155: Fyi, in Typescript when using the `length` property of an array to check if it's not empty, you can simply truthy check it and leave out the > 0.  But both work and it's a matter of style choice!

erdirowlands (approved):
🚀

src/__tests__/evaluator.test.ts
```diff
@@ -77,4 +85,157 @@ describe('Evaluator', () => {
...
+
+    await loadFlags(repository);
+    await loadSegments(repository);
+
+    const target = {
+      identifier: name,
+      name: name,
+      attributes: {
+        email: email,
+        role: role,
```
knagurski, line 228: You can short hand these
```suggestion
      name,
      attributes: {
        email,
        role,
```

src/__tests__/evaluator.test.ts
```diff
@@ -77,4 +85,157 @@ describe('Evaluator', () => {
...
+      },
+    };
+
+    const result = await evaluator.boolVariation(
+      flagName,
+      target,
+      !expected,
+      null,
```
knagurski, line 236: This is optional anyway, so you might as well omit it
```suggestion
```

src/__tests__/evaluator.test.ts
```diff
@@ -77,4 +85,157 @@ describe('Evaluator', () => {
...
+    ['email_is_mgr', 'boolflag_and', 'user@harness.io', 'manager', false],
+    ['ext_email_is_dev', 'boolflag_and', 'user@gmail.com', 'developer', false],
+    ['ext_email_is_mgr', 'boolflag_and', 'user@gmail.com', 'manager', false],
+    // if (target.attr.email endswith '@harness.io' || target.attr.email endswith '@somethingelse.com')
+    ['email_is_harness', 'boolflag_or', 'user@harness.io', 'n/a', true],
+    ['email_is_other', 'boolflag_or', 'user@somethingelse.com', 'n/a', true],
+    ['email_is_gmail', 'boolflag_or', 'user@gmail.com', 'n/a', false],
+  ])('Target rules v2 %s_%s', async (name, flagName, email, role, expected) => {
+    async function loadFlags(repo: StorageRepository) {
+      const feature_config_or: FeatureConfig = {
```
knagurski, line 101: Try to keep to standard JS naming conventions using camel case instead of snake case
```suggestion
      const featureConfigOr: FeatureConfig = {
```

src/__tests__/evaluator.test.ts
```diff
@@ -77,4 +85,157 @@ describe('Evaluator', () => {
...
+    // if (target.attr.email endswith '@harness.io' && target.attr.role = 'developer')
+    ['email_is_dev', 'boolflag_and', 'user@harness.io', 'developer', true],
+    ['email_is_mgr', 'boolflag_and', 'user@harness.io', 'manager', false],
+    ['ext_email_is_dev', 'boolflag_and', 'user@gmail.com', 'developer', false],
+    ['ext_email_is_mgr', 'boolflag_and', 'user@gmail.com', 'manager', false],
+    // if (target.attr.email endswith '@harness.io' || target.attr.email endswith '@somethingelse.com')
+    ['email_is_harness', 'boolflag_or', 'user@harness.io', 'n/a', true],
+    ['email_is_other', 'boolflag_or', 'user@somethingelse.com', 'n/a', true],
+    ['email_is_gmail', 'boolflag_or', 'user@gmail.com', 'n/a', false],
+  ])('Target rules v2 %s_%s', async (name, flagName, email, role, expected) => {
```
knagurski, line 99: Try to give the test human readable names so it's easier to track down without having to dig through the test code. If you want to have a section called `Target rules v2`, maybe consider adding a `describe` block around this.
```suggestion
  ])('it should do X when Y', async (name, flagName, email, role, expected) => {
```

You can play around with the ordering of the variables to get the `%s`s to line up properly, or add a short description of what you're trying to prove as the first element in the arrays and use that in the test name.

src/__tests__/evaluator.test.ts
knagurski, line 101: It's a good practice to extract mocking functions from the test into the `describe` block (or even out of the suit) so that reading the test is more straightforward.

src/evaluator.ts
knagurski, line 155: You can simplify a wee bit
```suggestion
        if (!!segment?.servingRules?.length) {
```

src/evaluator.ts
```diff
@@ -152,17 +152,31 @@ export class Evaluator {
...
+          // Use enhanced rules first if they're available
+          const sortedServingRules = segment.servingRules.sort(
```
knagurski, line 157: Just FYI, `sort` mutates the original array in place and returns itself. This will alter `segment.serviceRules`. 
Screenshot 2024-04-18 at 10 48 58

If you don't want to do that, you can spread the rules into a new array (or if we could target a newer version of Node use `toSorted` which does what you would expect rather than mutating).

```suggestion
          const sortedServingRules = [...segment.servingRules].sort(
```

Alternatively, if it doesn't matter that this is mutating the original array, then you can drop the `sortedServingRules` variable and just use the original.
```suggestion
          segment.servingRules.sort(
```

src/evaluator.ts
```diff
@@ -228,6 +242,20 @@ export class Evaluator {
...
+    target: Target,
+  ): Promise<boolean> {
+    for (const clause of clauses) {
+      if (!(await this.evaluateClause(clause, target))) {
+        // first clause to false, short-circuit and exit with false
+        return false;
+      }
+    }
+    // all clauses have passed
+    return true;
```
knagurski, line 256: What would happen if no clauses had been set up? Would this maybe cause the target to appear to be in the group erroneously?

knagurski (commented):
A few comments and things to think about

src/evaluator.ts
andybharness: that's a good point, we probably want that returning `false` if `clauses` is empty

src/evaluator.ts
andybharness: that's good to know about `sort()`, the other SDK are just mutating it in place. so i could probably get rid of that sorted variable at least

src/__tests__/evaluator.test.ts
andybharness: yeah i brought these over from Ruby, I probably left the old names in place

andybharness: yeah i brought these over from Ruby, I probably left the old names in place. fixed

src/evaluator.ts
andybharness: I'm going to leave off those `!!` because WebStrom is flagging that as redundant

image

knagurski: I'm too used to JSX and needing to make sure things are deffo boolean and not just truthy. Good they can be left out 👍🏻

PR #114, opened 2024-05-02
FFM-9097 Don't set global Axios timeout

erdirowlands:
# What
- No longer sets global timeout for Axios which would affect other instances of Axios in a user's application.  
- Creates a custom Axios instance setting options for SaaS or On-prem, with a user defined timeout from `Options`.  The timeout will default to `30000ms` if not set.

# Testing
Timeout configuration:
- Timeout not set by user defaults `30000ms`:
Screenshot 2024-05-02 at 14 05 57

- Timeout set to `40000ms` 
Screenshot 2024-05-02 at 14 05 36

Requests:
- All requests to SaaS work correctly: init/polling/streaming/metrics 
- On prem, not explictiy tested but loaded in CA file and observed that the custom Axios instance used it.

### DIFF
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,1 +1,1 @@
-export const VERSION = '1.7.0';
+export const VERSION = '1.8.0';
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,1 +1,1 @@
-export const VERSION = '1.7.0';
+export const VERSION = '1.9.0';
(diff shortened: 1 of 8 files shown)


src/version.ts
```diff
@@ -1 +1 @@
-export const VERSION = '1.7.0';
+export const VERSION = '1.8.0';
```
erdirowlands, line 1: The version was bumped to `1.8.0` in the last PR and has not been released yet.  We forgot to update the version here in that PR.

andybharness (approved):

PR #115, opened 2024-05-02
FFM-11364  Use target-segments v2-rule parameter

erdirowlands:
# What
- Re-generates the API client to bring in the new target-segments query parameter rules
- Sets this parameter on the for requests to to `target-segments` (polling) and  `target-segments/{segment}` (streamed group changes)

# Testing

Manual inspection of final URL string when
- Pulling all segments
Screenshot 2024-05-02 at 20 05 21

- Pulling a specific segment when a group change has been streamed
Screenshot 2024-05-02 at 20 04 51

### DIFF
--- a/src/streaming.ts
+++ b/src/streaming.ts
@@ -18,6 +18,7 @@ type FetchFunction = (
   identifier: string,
   environment: string,
   cluster: string,
+  // Only target-segment requests require this, and it will be safely ignored for feature calls.
   targetSegmentRulesQueryParameter: string,
 ) => AxiosPromise<FeatureConfig | Segment>;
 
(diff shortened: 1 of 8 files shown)


src/streaming.ts
```diff
@@ -18,6 +18,8 @@ type FetchFunction = (
   identifier: string,
   environment: string,
   cluster: string,
+  // Only target-segment requests require this, and it will be safely ignored for feature calls.
+  targetSegmentRulesQueryParameter: string,
```
erdirowlands, line 22: It's not perfect, having this argument "ignored" by featureByIdentifier requests, but with the way this class is setup with a generic style message processor, it was the only way without a re-write.  I've confirmed using the debugger that this won't get passed in the eventual request to featureByIdentifier.

PR #122, opened 2025-03-19
FFM-12353 Reduce excessive logs

erdirowlands:
# What
Reduces log spam during stream connection issues by reducing frequency and level of logs depending on attempt #:

- Error log for `Issue with streaming: falling back to polling while the SDK attempts to reconnect`: Only show once per disconnect cycle

- Warning logs: Show on 1st, 10th, and every 50th retry attempt, with debug-level in between

- Combines disconnect and retry message in one log.

- Other reductions of logs which could become spammy. 

# Testing
Verified in local environment with unstable connections that logs appear at expected thresholds and reconnection resets counters correctly.

### DIFF
--- a/src/sdk_codes.ts
+++ b/src/sdk_codes.ts
@@ -160,7 +160,7 @@ export function warnAuthRetrying(
 
 // Track disconnect attempts for throttling warning logs
 let disconnectAttempts = 0;
-let lastConnectionSuccess = 0;
+let lastConnectionSuccess = Date.now(); // Initialize with current time to avoid huge numbers
 
 // Restart disconnect counter when connection succeeds
 export function restartDisconnectCounter(): void {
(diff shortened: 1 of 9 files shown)


src/sdk_codes.ts
```diff
@@ -158,14 +158,53 @@ export function warnAuthRetrying(
...
-  logger.warn(getSdkErrMsg(5003, `${seconds}`));
+// Reset disconnect counter
+export function resetDisconnectCounter(): void {
+  disconnectAttempts = 0;
+  lastConnectionSuccess = 0;
 }
 
+
+// Combined warning function for both disconnect and retry in a single message
+export function warnStreamDisconnectedWithRetry(reason: string, ms: number, logger: Logger): void {
```
erdirowlands, line 179: Arguable that this logic should not be encapsulated in sdk_codes, but I think it's a small enough change as to not cause maintainability issues.

jcox250 (approved):

PR #123, opened 2025-04-03
FFM-12374 Retry enhancements

erdirowlands:
# What
Retries
- Now retries on `5xx` errors for auth requests (auth was previously excluded because POST isn’t considered idempotent).
- Now retries on Axios timeouts (ECONNABORTED).
- Cleaned up error logs to avoid dumping entire Axios error objects

SDK Init
- SDK is now only marked “initialized” (and emits READY) after successfully fetching flags/segments. This ensures waitForInitialization() blocks until flags and segments are actually laoded. 

Dependency upgrades
- Bumps `axios-retry` to get `onMaxRetryTimesExceeded` - this is compatible with our min supported Node `12` version.
- Bumps `esbuild` due to a medium sev.  Note the SDK now requires Node v18 to build, but as this is a dev dependency it doesn't affect users.

# Testing
Manual and TestGrid integration app. Node versions 12 / 16/ 18

### DIFF
--- a/src/polling.ts
+++ b/src/polling.ts
@@ -105,7 +105,9 @@ export class PollingProcessor {
         this.repository.setFlag(fc.feature, fc),
       );
     } catch (error) {
-      this.log.error('Error loading flags', error);
+      this.log.error(
+        `Error loading flags (${error.code ?? "UNKNOWN"}): ${error.message}`
+      );
       throw error;
     }
   }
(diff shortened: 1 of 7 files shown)


src/polling.ts
```diff
@@ -80,7 +80,7 @@ export class PollingProcessor {
         }
       })
       .catch((error) => {
-        this.eventBus.emit(PollerEvent.ERROR, { error });
```
erdirowlands, line 83: No need to wrap the error in an object.

conormurray95 (approved):

